PR #1, opened 2019-05-17
PP-5219 add logging framework to project

SandorArpa:
## WHAT
- add logging framework
- add /v1/ to resource path


src/main/resources/config/config.yaml
```diff
@@ -0,0 +1,8 @@
+logging:
```
mrlumbu: maybe we should think about the logging format, and whether we can use JSON logging straight away in this app?

PR #4, opened 2019-05-24
PP-5115: Add Dockerfile and startup scripts

rjbaker:
### WHAT
Adds a simple openjdk11 alpine container, startup script and local build script similar to other Pay application services.

### HOW
This will allow a local application container to be built and run (requires local installation of maven). Once merged, further build pipeline development may continue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,29 @@
+# adoptopenjdk/openjdk11:jre-11.0.3_7-alpine
+FROM adoptopenjdk/openjdk11@sha256:eaa182283f19d3f0ee0c6217d29e299bb4056d379244ce957e30dcdc9e278e1e
+
+RUN ["apk", "--no-cache", "upgrade"]
+
+ARG DNS_TTL=15
+
+# Default to UTF-8 file.encoding
+ENV LANG C.UTF-8
+
+RUN echo networkaddress.cache.ttl=$DNS_TTL >> "$JAVA_HOME/conf/security/java.security"
+
+# Add RDS CA certificates to the default truststore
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+RUN wget -qO - https://s3.amazonaws.com/rds-downloads/rds-ca-2015-root.pem       | keytool -importcert -noprompt -cacerts -storepass changeit -alias rds-ca-2015-root \
+ && wget -qO - https://s3.amazonaws.com/rds-downloads/rds-combined-ca-bundle.pem | keytool -importcert -noprompt -cacerts -storepass changeit -alias rds-combined-ca-bundle
+
+RUN ["apk", "add", "--no-cache", "bash"]
+
+ENV PORT 8080
+ENV ADMIN_PORT 8081
+
+EXPOSE 8080
+EXPOSE 8081
```
rjbaker: Is the admin port required by this application?

mrlumbu: No

docker-startup.sh
```diff
@@ -0,0 +1,29 @@
...
+  i=0
+  truststore_pass=changeit
+  for cert in "$CERTS_PATH"/*; do
+    [ -f "$cert" ] || continue
+    echo "Adding $cert to default truststore"
+    keytool -importcert -noprompt -cacerts -storepass "$truststore_pass" -file "$cert" -alias custom$((i++))
+  done
+fi
+
+java $JAVA_OPTS -jar *-allinone.jar waitOnDependencies *.yaml
```
mrlumbu: we don't have this yet, and maybe this isn't the best way to do it. Could maybe just use db status check provided by dropwizard? https://www.dropwizard.io/0.7.1/docs/manual/migrations.html

rjbaker: I'll comment this out for now, but feel free to amend as you continue development. This should be enough container setup to get the Jenkins build pipeline running.

mrlumbu: 👍

PR #5, opened 2019-05-24
PP-5115: Add ledger Jenkinsfile build configuration

rjbaker:
### WHAT:
Adds a Jenkinsfile to build, test and (later) deploy the ledger service.

### How to test & deploy:
The Jenkins CI configuration must be updated in alphagov/pay-chef to scan and build this repository. Once this has been deployed to CI, this application may be built.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,143 @@
+#!/usr/bin/env groovy
+
+pipeline {
+  agent any
+
+  parameters {
+    booleanParam(defaultValue: false, description: '', name: 'runEndToEndTestsOnPR')
+  }
+
+  options {
+    ansiColor('xterm')
+    timestamps()
+  }
+
+  libraries {
+    lib("pay-jenkins-library@master")
+  }
+
+  environment {
+    DOCKER_HOST = "unix:///var/run/docker.sock"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,143 @@
...
+          }
+        }
+      }
+      post {
+        failure {
+          postMetric("ledger.docker-build.failure", 1)
+        }
+      }
+    }
+    stage('Tests') {
```
mrlumbu: we don't need any e2e tests to start with as none of our current e2e tests will touch this service. We can change this as and when we write those tests

mrlumbu: i.e. you can remove this whole stage for now

rjbaker: Thanks. No simple way to skip. Have removed for now. 👍

Jenkinsfile
```diff
@@ -0,0 +1,127 @@
...
+        }
+      }
+    }
+    stage('Deploy') {
+      when {
+        branch 'master'
+      }
+      steps {
+        checkPactCompatibility("ledger", gitCommit(), "test")
+        deployEcs("ledger")
```
whpearson: This will attempt to deploy to ecs, so will fail for now.

PR #15, opened 2019-06-04
PP-5250 create event table

SandorArpa:
## WHAT
- add liquibase create event table
- add insert method to EventDao with ResourceType.id lookup
- add EventFixture
- add relevant tests
- ignore integration tests until further notice

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/model/Event.java
+++ b/src/main/java/uk/gov/pay/ledger/event/model/Event.java
@@ -1,19 +1,64 @@
 package uk.gov.pay.ledger.event.model;
 
-import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import uk.gov.pay.ledger.event.model.serializer.MicrosecondPrecisionDateTimeSerializer;
 
+import java.time.ZonedDateTime;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
 public class Event {
-    private String id;
 
-    public Event() {
-    }
+    @JsonIgnore
+    private Long id;
+    private String sqsMessageId;
+    private ResourceType resourceType;
+    private String resourceExternalId;
+    @JsonSerialize(using = MicrosecondPrecisionDateTimeSerializer.class)
+    private ZonedDateTime eventDate;
+    private String eventType;
+    private String eventData;
 
-    public Event(String id) {
+    public Event() { }
+
+    public Event(Long id, String sqsMessageId, ResourceType resourceType, String resourceExternalId,
+                 ZonedDateTime eventDate, String eventType, String eventData) {
         this.id = id;
+        this.sqsMessageId = sqsMessageId;
+        this.resourceType = resourceType;
+        this.resourceExternalId = resourceExternalId;
+        this.eventDate = eventDate;
+        this.eventType = eventType;
+        this.eventData = eventData;
     }
 
-    @JsonProperty
-    public String getId() {
+    public Long getId() {
         return id;
+    }
+
+    public String getSqsMessageId() {
+        return sqsMessageId;
+    }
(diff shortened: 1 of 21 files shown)


src/main/resources/migrations/00001_create_table_event.sql
```diff
@@ -2,6 +2,19 @@
 
 --changeset uk.gov.pay:build event table
 CREATE TABLE event (
-    id VARCHAR(255) PRIMARY KEY
+    id BIGSERIAL PRIMARY KEY,
```
kbottla: Event structure envisioned here is different from https://github.com/alphagov/pay-architecture/blob/master/reporting/domain_model.md

heathd: suggestion after discussion with @mrlumbu @kbottla @SandorArpa 

```
CREATE TABLE event (
    id BIGSERIAL PRIMARY KEY,
    sqs_message_id VARCHAR(255) NOT NULL,
    resource_type INT FOREIGN KEY RESOURCES(ID),
    resource_external_id VARCHAR(255),
    event_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL
    event_type VARCHAR(255),
    event_data jsonb not null,
);
```

src/main/resources/migrations/00003_create_table_event.sql
```diff
@@ -0,0 +1,26 @@
...
+--changeset uk.gov.pay:add_event_resource_type_id_fk
+ALTER TABLE event ADD CONSTRAINT event_resource_type_id_fk FOREIGN KEY (resource_type_id) REFERENCES resource_type (id);
+--rollback drop constraint event_resource_type_id_fk;
+
+--changeset uk.gov.pay:add_index_resource_external_id
+CREATE INDEX index_resource_external_idx ON event(resource_external_id);
+--rollback drop index index_resource_external_idx;
+
+--changeset uk.gov.pay:add_index_event_type
+CREATE INDEX index_event_typex ON event(event_type);
```
kbottla: Can we defer index on `event_type` ? Do see it being used in any queries

src/main/java/uk/gov/pay/ledger/event/model/Event.java
```diff
@@ -1,19 +1,64 @@
...
-    private String id;
 
-    public Event() {
-    }
+    @JsonIgnore
+    private Long id;
+    private String sqsMessageId;
+    private ResourceType resourceType;
+    private String resourceExternalId;
+    @JsonSerialize(using = ApiResponseDateTimeSerializer.class)
```
heathd: not a biggie, but we're storing [microsecond precision](https://github.com/alphagov/pay-connector/blob/master/src/main/java/uk/gov/pay/connector/events/MicrosecondPrecisionDateTimeSerializer.java) so probably we should emit that

src/test/java/uk/gov/pay/ledger/event/EventDaoITest.java
```diff
@@ -0,0 +1,46 @@
...
+    @ClassRule
+    public static AppWithPostgresRule rule = new AppWithPostgresRule();
+
+    @Test
+    public void shouldInsertEvent() {
+        Event event = anEventFixture().toEntity();
+        EventDao eventDao = rule.getJdbi().onDemand(EventDao.class);
+        eventDao.insertEventWithResourceTypeId(event);
+        DatabaseTestHelper dbHelper = aDatabaseTestHelper(rule.getJdbi());
+        Map<String, Object> result = dbHelper.getEventByExternalId(event.getResourceExternalId());
```
kbottla: Can we assert the DB columns/fields match the expected ?

src/test/java/uk/gov/pay/ledger/event/EventDaoITest.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Test
+    public void shouldFindEvent() {
+        Event event = anEventFixture()
+                .insert(rule.getJdbi())
+                .toEntity();
+        EventDao eventDao = rule.getJdbi().onDemand(EventDao.class);
+        Optional<Event> optionalEvent = eventDao.getById(event.getId());
+        assertThat(optionalEvent.isPresent(), is(true));
```
kbottla: Same here for asserting columns

PR #23, opened 2019-06-07
PP-5256 Added TransactionViewDao

kbottla:
## WHAT 

- Added TransactionViewDao (`searchTransactionView()`) to support searching for transactions (to be used by Transaction search resource endpoint). 
- Supports search by fields available for a `PaymentCreated` event 
- Fields currently supported are 
` accountId, email, reference, cardHolderName, fromDate, toDate, pageNumber, displaySize, state`
-  No search results are returned if the following parameters are set as these are not available yet
` refundStates, lastDigitsCardNumber,  firstDigitsCardNumber, cardBrands`
- Results are ordered by `Transaction.id` but to be replaced by transaction_date/similar when available

- TransactionViewDao is added to support cusom query instead of existing `TransactionDao` which is an interface and leverages JDBI bindings

with @SandorArpa

## HOW
- `TransactionViewDaoITest` should pass in local dev environment. Currently ignored due to postgres container startup issue in CI

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/common/TransactionState.java
+++ b/src/main/java/uk/gov/pay/ledger/common/TransactionState.java
@@ -1,0 +1,41 @@
+package uk.gov.pay.ledger.common;
+
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum TransactionState {
+
+    CREATED("created", false);
+
+    private final String value;
+    private final boolean finished;
+    private final String code;
+    private final String message;
+
+    TransactionState(String value, boolean finished) {
+        this.value = value;
+        this.finished = finished;
+        this.code = null;
+        this.message = null;
+    }
+
+    @JsonProperty("status")
+    public String getState() {
+        return value;
+    }
+
+    public boolean isFinished() {
+        return finished;
+    }
+
+    public String getCode() {
+        return code;
+    }
+
+    public String getMessage() {
+        return message;
+    }
+}
--- a/src/main/java/uk/gov/pay/ledger/transaction/TransactionSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/TransactionSearchParams.java
@@ -1,0 +1,180 @@
+package uk.gov.pay.ledger.transaction;
+
+import java.time.ZonedDateTime;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import static org.apache.commons.lang3.StringUtils.isNotBlank;
+
+public class TransactionSearchParams {
+    private static final String GATEWAY_ACCOUNT_EXTERNAL_FIELD = "gatewayAccountExternalId";
+    private static final String OFFSET_FIELD = "offset";
+    private static final String PAGE_SIZE_FIELD = "limit";
+    private static final String CARDHOLDER_NAME_FIELD = "cardholderName";
+    private static final String FROM_DATE_FIELD = "fromDate";
+    private static final String TO_DATE_FIELD = "toDate";
+    private static final String EMAIL_FIELD = "email";
+    private static final String REFERENCE_FIELD = "reference";
+    private static final long MAX_DISPLAY_SIZE = 500;
+
+    private String accountId;
+    private String email;
+    private String reference;
+    private String cardHolderName;
+    private String lastDigitsCardNumber;
+    private String firstDigitsCardNumber;
+    private CommaDelimitedSetParameter paymentStates;
+    private CommaDelimitedSetParameter refundStates;
+    private List<String> cardBrands;
+    private ZonedDateTime fromDate;
+    private ZonedDateTime toDate;
+    private Long pageNumber;
+    private Long displaySize;
+
+    private Map<String, Object> queryMap;
+
+    public void setAccountId(String accountId) {
+        this.accountId = accountId;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+
+    public void setReference(String reference) {
+        this.reference = reference;
+    }
+
+    public void setCardHolderName(String cardHolderName) {
+        this.cardHolderName = cardHolderName;
+    }
+
+    public void setLastDigitsCardNumber(String lastDigitsCardNumber) {
+        this.lastDigitsCardNumber = lastDigitsCardNumber;
+    }
+
+    public void setFirstDigitsCardNumber(String firstDigitsCardNumber) {
+        this.firstDigitsCardNumber = firstDigitsCardNumber;
+    }
+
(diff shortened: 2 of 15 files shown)


mrlumbu:
Maybe a bit more detail on what you've done and why? - one line of comment for 850 lines of code seems a bit on the low side 😃

kbottla:
@mrlumbu Was updating comment and is done now

src/test/java/uk/gov/pay/ledger/transaction/TransactionViewDaoITest.java
```diff
@@ -0,0 +1,377 @@
...
+        TransactionSearchParams searchParams = new TransactionSearchParams();
+        searchParams.setAccountId(gatewayAccountId);
+        searchParams.setDisplaySize(2l);
+        searchParams.setPageNumber(3l);
+
+
+        List<TransactionView> viewList = transactionViewDao.searchTransactionView(searchParams);
+
+        assertThat(viewList.size(), Matchers.is(2));
+        assertThat(viewList.get(0).getReference(), is("reference15"));
```
mrlumbu: why is this reference 15 on page 3 with page size of 2? I'm probably just being dumb, but can't see how that has worked.

src/main/java/uk/gov/pay/ledger/common/TransactionState.java
```diff
@@ -0,0 +1,41 @@
...
+    @JsonProperty("status")
+    public String getState() {
+        return value;
+    }
+
+    public boolean isFinished() {
+        return finished;
+    }
+
+    public String getCode() {
```
mrlumbu: Not sure what the point of this getter is

src/main/java/uk/gov/pay/ledger/common/TransactionState.java
```diff
@@ -0,0 +1,41 @@
+package uk.gov.pay.ledger.common;
```
mrlumbu: think this should go in transaction package unless there is a good reason not to

src/main/java/uk/gov/pay/ledger/transaction/TransactionSearchParams.java
```diff
@@ -0,0 +1,180 @@
+package uk.gov.pay.ledger.transaction;
```
mrlumbu: please can you put each of your new classes in an appropriate sub package, or even put everything to do with search in a new package

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionViewDao.java
```diff
@@ -0,0 +1,37 @@
...
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.statement.Query;
+import uk.gov.pay.ledger.transaction.TransactionSearchParams;
+import uk.gov.pay.ledger.transaction.TransactionView;
+import uk.gov.pay.ledger.transaction.dao.mapper.TransactionViewMapper;
+
+import javax.inject.Inject;
+import java.util.List;
+
+public class TransactionViewDao {
```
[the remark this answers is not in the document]
kbottla: Updated to `TransactionSearchDao` but we can revisit if needed in the future

src/test/java/uk/gov/pay/ledger/transaction/TransactionViewDaoITest.java
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    @Test
+    public void shouldReturn2Records_whenSearchingBySpecificGatewayAccountId() {
+        String gatewayAccountId = "account-id-" + nextLong();
+
+        for (int i = 0; i < 2; i++) {
+            aTransactionFixture()
+                    .withGatewayAccountId(gatewayAccountId)
+                    .withAmount(100l + i)
```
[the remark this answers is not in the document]
kbottla: 👍 Removed

src/test/java/uk/gov/pay/ledger/transaction/TransactionViewDaoITest.java
kbottla: This is because search orders results by ID `desc`.  Transaction IDs will start from xxx19 to xxx1.

So for display size of 2
     page 1 should return xxx19, xxx18
     page 2 should return xxx17, xxx16
     page 3 should return xxx15, xxx14

src/test/java/uk/gov/pay/ledger/transaction/TransactionViewDaoITest.java
```diff
@@ -0,0 +1,377 @@
...
+
+        List<TransactionView> viewList = transactionViewDao.searchTransactionView(searchParams);
+
+        assertThat(viewList.size(), Matchers.is(2));
+        assertThat(viewList.get(0).getGatewayAccountId(), is(gatewayAccountId));
+        assertThat(viewList.get(1).getGatewayAccountId(), is(gatewayAccountId));
+    }
+
+    @Test
+    public void shouldReturn1Record_whenSearchingByEmail() {
```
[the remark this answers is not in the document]
kbottla: We think `FilterByEmail` is being tested by the number of records we are expecting and `whenSearchingByEmail` conveys the same. Don't see the current name is ambiguous. Happy to have a chat

src/main/java/uk/gov/pay/ledger/transaction/TransactionView.java
```diff
@@ -0,0 +1,128 @@
...
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import uk.gov.pay.commons.api.json.ApiResponseDateTimeSerializer;
+import uk.gov.pay.ledger.common.TransactionState;
+import uk.gov.pay.ledger.transaction.model.CardDetails;
+
+import java.time.ZonedDateTime;
+import java.util.Objects;
+
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
+public class TransactionView {
```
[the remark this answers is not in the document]
mrlumbu: Ok I see what you mean, but I think if we want a view or representation class, it should wrap what we get from the database, it shouldn't actually _be_ what we get from the db. What you are getting from the db is `List<Transaction>`, so I think you should use the `Transaction` class. A resource or service can then add the necessary representational bits.

src/test/java/uk/gov/pay/ledger/transaction/TransactionViewDaoITest.java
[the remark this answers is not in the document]
mrlumbu: Yeah, I'm just being pedantic 😄 
I guess if a test says `shouldReturn1ofX`, I'm expecting the thing it's testing to always return 1 thing (e.g. it returns the head of a list). Whereas this is just testing that searching by email returns a list of things matching that email, so I think the test name should reflect that.

src/test/java/uk/gov/pay/ledger/transaction/TransactionViewDaoITest.java
mrlumbu: I see

mrlumbu:
I do think we need to think more about the TransactionView thing, but I'm happy for that to be done in a follow up PR. Let's discuss if it's not clear what I'm on about

PR #27, opened 2019-06-11
PP-5256 Transaction Search - Adds links

kbottla:
## WHAT 

- Added pagination links for transaction search results
- Added self, refund links for transactions
- Unit tests - PaginationBuilderTest, TransactionServiceTest (for search results)

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/model/Link.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/model/Link.java
@@ -1,0 +1,60 @@
+package uk.gov.pay.ledger.transaction.search.model;
+
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+
+import java.util.Objects;
+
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/ledger/transaction/search/model/Link.java
```diff
@@ -0,0 +1,59 @@
...
+                '}';
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+        Link link = (Link) o;
+        return Objects.equals(href, link.href) &&
+                Objects.equals(method, link.method);
```
SandorArpa: Maybe we could include `Objects.equals(rel, link.rel)` as well? 🤔

kbottla: 👍 Updated

PR #28, opened 2019-06-12
PP-5256 Transaction search - resource endpoint

kbottla:
## WHAT 
- Adds resource endpoint `/v1/api/accounts/{accountId}/transactions` to search for transactions
- Integration test for search
- Updated `/v1/transaction` to `/accounts/{accountId}/transaction/{transactionExternalId}` and relevant changes


src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -13,7 +13,7 @@
 public class TransactionDao {
 
     private static final String FIND_TRANSACTION_BY_EXTERNAL_ID = "SELECT * FROM transaction " +
-            "WHERE external_id = :externalId";
+            "WHERE external_id = :externalId and gateway_account_id= :gatewayAccountId ";
```
SandorArpa: Maybe we want to switch it around? 🤔 Eg, `WHERE gateway_account_id= :gatewayAccountId AND external_id = :externalId`
Reason: method that's using this query has its parameters in this order pedantry)

kbottla: We will update it in the next if you would like

PR #30, opened 2019-06-13
PP-5254 match CardDetails's Json format to that of connector

SandorArpa:
## WHAT
- match the Json format of `CardDetails` to connector


SandorArpa:
wrong Jira ticket number 🤕

PR #37, opened 2019-06-17
PP-5257 Change API structure for transactions

kbottla:
## WHAT
- Transaction API endpoints changed from `/v1/api/accounts/{accountId}....` to
        `/v1/transaction/{transactionExternalId}` - to find a single transaction
        `/v1/transaction?query_params` - to search transactions
- Validation for `accound_id` which is now a query parameter for transaction search endpoint
- Updates relevents DAO queries, methods, and tests to reflect above changes

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
@@ -1,18 +1,26 @@
 package uk.gov.pay.ledger.transaction.search.common;
 
 import uk.gov.pay.ledger.exception.UnparsableDateException;
+import uk.gov.pay.ledger.exception.ValidationException;
 
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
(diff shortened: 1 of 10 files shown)


src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
```diff
@@ -1,18 +1,26 @@
...
 
 public class TransactionSearchParamsValidator {
 
+    private static final String GATEWAY_ACCOUNT_ID = "account_id";
     private static final String FROM_DATE_FIELD = "from_date";
     private static final String TO_DATE_FIELD = "to_date";
 
     public static void validateSearchParams(TransactionSearchParams searchParams) {
+
+        throwIfBlankFieldValue(GATEWAY_ACCOUNT_ID, searchParams.getAccountId());
```
jankowiakmaria: Do we want to always enforce this value?

kbottla: Yes, it is always required for now

PR #39, opened 2019-06-17
PP-5255 Add EventType

mrlumbu:
Introduce EventType enum. As part of this I have changed the name of the expected event from PaymentCreated to PAYMENT_CREATED. This is because whilst Jackson is pretty good at {,de}serialisation, JDBI is not so good, and it's default enum deserialiser uses the enum `name()`, so it
significantly reduces boiler plate to use this form.


jankowiakmaria:
I believe this PR can be closed. The changes have been merged as part of #43

PR #43, opened 2019-06-18
Pp 5255 event digest

mrlumbu:
First step in going from receiving an event to writing a transaction. The main idea is the event digest, 
that is used to make sense of the event history for a given resource_external_id.

The EventDigest is an object that represents the summary of a list of events pertinent to the further processing of those events. Basically that amounts to the timestamp and event type of the latest event by timestamp, as well as a compressed digest of the event details attached to each event. The idea is that this EventDigest is easily convertible into a Transaction. The way event details are compressed, if two event_detail objects both contain an attribute then the later event will overwrite the earlier one.  

This PR also introduces EventType enum. As part of this I have changed the name of the expected event from PaymentCreated to PAYMENT_CREATED. This is because whilst Jackson is pretty good at {,de}serialisation, JDBI is not so good, and it's default enum deserialiser uses the enum `name()`, so it significantly reduces boiler plate to use this form.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/dao/EventDao.java
+++ b/src/main/java/uk/gov/pay/ledger/event/dao/EventDao.java
@@ -11,6 +11,7 @@ import org.jdbi.v3.sqlobject.transaction.Transaction;
 import uk.gov.pay.ledger.event.dao.mapper.EventMapper;
 import uk.gov.pay.ledger.event.model.Event;
 
+import java.util.List;
 import java.util.Optional;
 
 @RegisterRowMapper(EventMapper.class)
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/ledger/event/dao/EventDao.java
```diff
@@ -23,6 +24,11 @@
             " FROM event e, resource_type rt WHERE e.id = :eventId AND e.resource_type_id = rt.id")
     Optional<Event> getById(@Bind("eventId") Long eventId);
 
+    @SqlQuery("SELECT e.id, e.sqs_message_id, rt.name AS resource_type_name, e.resource_external_id, e.event_date," +
+            " e.event_type, e.event_data" +
+            " FROM event e, resource_type rt WHERE e.resource_external_id = :resourceExternalId AND e.resource_type_id = rt.id")
+    Optional<Event> getByResourceExternalId(@Bind("resourceExternalId") String resourceExternalId);
```
jankowiakmaria: I can't see it being used anywhere. Am I missing something?

mrlumbu: @jankowiakmaria I have removed this

mrlumbu:
@jankowiakmaria thanks, addressed your comments

PR #45, opened 2019-06-18
BAU: fix flaky tests

jankowiakmaria:
* There was some sort of interference between the tests that caused sqs integration tests to fail when run with all of them
* Introduced `backgroundProcessingEnabled` configuration variable which is turned off for all the tests but selected one
* small cleanup in `EventQueueIT` test

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/rule/AppWithPostgresAndSqsRule.java
+++ b/src/test/java/uk/gov/pay/ledger/rule/AppWithPostgresAndSqsRule.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.ledger.rule;
 
 import com.amazonaws.services.sqs.AmazonSQS;
+import io.dropwizard.testing.ConfigOverride;
 import io.dropwizard.testing.junit.DropwizardAppRule;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.sqlobject.SqlObjectPlugin;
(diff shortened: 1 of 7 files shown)


src/test/java/uk/gov/pay/ledger/rule/AppWithPostgresAndSqsRule.java
```diff
@@ -18,22 +18,30 @@
...
-    private final AmazonSQS sqsClient;
+    private AmazonSQS sqsClient;
     private PostgreSQLContainer postgres;
     private DropwizardAppRule<LedgerConfig> appRule;
 
     public AppWithPostgresAndSqsRule() {
+        this(false);
+    }
+
+    public AppWithPostgresAndSqsRule(boolean enableBackgroundProcessing) {
```
kbottla: We could use config override which give us flexibility to override any variable. See example
https://github.com/alphagov/pay-connector/blob/84860a0eb8bd364fd280b4a6440f4792e9bfc6ff/src/test/java/uk/gov/pay/connector/rules/AppWithPostgresRule.java#L37

jankowiakmaria: good suggestion

jankowiakmaria: Updated

PR #47, opened 2019-06-18
BAU - reorganise test classes

kbottla:
## WHAT 
- Reorganise test classes (including integration tests) to mirror packages
- Updated few integration tests suffix to IT so fail safe plugin can run them as integration tests


src/test/java/uk/gov/pay/ledger/queue/QueueMessageReceiverIT.java
```diff
@@ -1,4 +1,4 @@
-package uk.gov.pay.ledger.it.queue.sqs;
+package uk.gov.pay.ledger.queue;
```
jankowiakmaria: [Maybe] `QueueMessageReceiver` is under `uk.gov.pay.ledger.queue.managed` so maybe this test should go there too.

PR #60, opened 2019-06-21
PP-5255 Insert transaction on recieving event

mrlumbu:
Adds the connecting logic that takes an EventDigest and creates and
inserts a transaction.

At the moment this simply inserts a transaction, so will not handle multiple
events. It also hardcodes the state of the transaction, so the mapping
logic needs writing.

In order to do this I have dropped some constraints on the transaction table.
This is because we don;t know what order we are going to recieve events,
so we may have only a minimal amount of information to create the
transaction, therefore we cannot enforce non-null constraints on things like
description, reference etc.

Another mildly controversial thing I have done is equate the event
`resourceExternalId` with the transaction externalId. I think this
is correct, as ultimately we need to be able to retrieve things from
connector and ledger with the same id.

This commit also does not contain any db transactions around
handling the event and creating transaction. I have thought about
this, but will hold off worrying about it for another commit.

### DIFF
--- a/src/main/resources/migrations/00005_drop_not_null_transaction_table.sql
+++ b/src/main/resources/migrations/00005_drop_not_null_transaction_table.sql
@@ -1,0 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:modify_transaction_table_drop_not_null_constraints
+
+ALTER TABLE transaction ALTER COLUMN gateway_account_id DROP NOT NULL;
+ALTER TABLE transaction ALTER COLUMN amount DROP NOT NULL;
+ALTER TABLE transaction ALTER COLUMN reference DROP NOT NULL;
+ALTER TABLE transaction ALTER COLUMN description DROP NOT NULL;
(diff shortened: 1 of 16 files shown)


heathd:
> Another mildly controversial thing I have done is equate the event resourceExternalId with the transaction externalId

this seems perfectly reasonable to me. The semantics of transaction externalId is down to us to define, and that's exactly what I would expect to do (a transaction representing a refund would have the externalId of the refund)

src/main/resources/migrations/00005_drop_not_null_transaction_table.sql
```diff
@@ -0,0 +1,4 @@
+ALTER TABLE transaction ALTER COLUMN gateway_account_id DROP NOT NULL;
```
SandorArpa: Can we add the usual liquibase markings, eg `--liquibase formatted sql` and assign a liquibase changeset id eg `--changeset uk.gov.pay:drop_not_null_transaction_table` ? I've seen unmarked files wreck havoc when changeset id is missing.

mrlumbu: Thanks, fixed

PR #64, opened 2019-06-24
Pp 5357 consume authorise success event

mrlumbu:
Removes EventType enum in favour of SalientEventType. There is now no validation of event type on reception of event, so emitters can send any event type they want. The SalientEventType enum will be used to translated a known set of events to a known set of transaction states. This is done via the EventDigest, which now stores the 'latestSalientEventType'.

Introduces handling of multiple different events concerning a single external resource id. This comprises:
- Changing TransactionService to map from latest salient event to transaction state
- Changing TransactionDao to upsert rather than insert transactions, using `ON CONFLICT DO UPDATE …`
- Introduce event count to avoid stale updates of transaction - transaction will only be updated if the changes are built of more events than current record
- introduces new transaction state 'SUBMITTED', and maps it to AUTHORISATION SUCCESS event


src/test/java/uk/gov/pay/ledger/pact/TransactionsApiContractTest.java
```diff
@@ -56,7 +56,7 @@ public void createTransactionWithCardDetails(Map<String, String> params) {
                 .withAmount(1000l)
                 .withReference("aReference")
                 .withDescription("Test description")
-                .withState("created")
+                .withState("CREATED")
```
heathd: I *think* this should use `TransactionState` rather than a string?

heathd:
I think that by "Salient" event you mean one which causes a state change. 

I wonder if it might be more self-evident if we just called it "StateChangingEvent"?

mrlumbu:
> I think that by "Salient" event you mean one which causes a state change.
> 
> I wonder if it might be more self-evident if we just called it "StateChangingEvent"?

Heh, maybe. I think we've been talking about salient events for so long I lost sight of the fact it's a bit of a quirky name

heathd:
also something to keep in mind.. the unique constraint will fail when deploying if there are dups (in test env)

heathd:
I'd consider renaming `SalientEvent` to `StateChangingEvent` but not a blocker to merging

PR #69, opened 2019-06-26
PP-5355 Example message consumer pact test

mrlumbu:
Example of how Ledger can generate a message pact for a particular event type. I think this can be improved, but this shows the basic idea of how to generate and verify a message pact for the `PAYMENT_CREATED` event.
At the moment the verification step checks that the event has been persisted to the database. I think this might be the right point to verify. We would then rely on the pacts and test fixtures being generated from a shared source to ensure end to end compatibility.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,12 +24,23 @@ pipeline {
 
   stages {
     stage('Maven Build') {
+      when {
+        branch 'master'
+      }
       steps {
         script {
-          long stepBuildTime = System.currentTimeMillis()
-          sh 'mvn -version'
-          sh 'mvn clean verify'
-          runProviderContractTests()
+          def stepBuildTime = System.currentTimeMillis()
+          def commit = gitCommit()
+          def branchName = 'master'
+
+          withCredentials([
+                  string(credentialsId: 'pact_broker_username', variable: 'PACT_BROKER_USERNAME'),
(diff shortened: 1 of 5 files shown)


oswaldquek:
wooo 👏

jankowiakmaria:
Does it need to be rebased?

Jenkinsfile
```diff
@@ -39,6 +50,35 @@ pipeline {
...
+          def stepBuildTime = System.currentTimeMillis()
+          def commit = gitCommit()
+          def branchName = gitBranchName()
+
+          withCredentials([
+                  string(credentialsId: 'pact_broker_username', variable: 'PACT_BROKER_USERNAME'),
+                  string(credentialsId: 'pact_broker_password', variable: 'PACT_BROKER_PASSWORD')]
+          ) {
+              sh 'mvn -version'
+              sh "mvn clean package pact:publish -DrunContractTests -DPACT_BROKER_URL=https://pact-broker-test.cloudapps.digital -DPACT_CONSUMER_VERSION=${commit}" +
```
jankowiakmaria: It seems the only difference between master and branch build is `runContractTests` parameter.
Is it not possible to aggregate them into one?

mrlumbu: Good point, but as it stands achieving this seemingly simple task is weirdly hard. It should be done, but I suggest we merge this as is to unblock further improvements of pact stuff

PR #76, opened 2019-06-28
PP-5357 Pass through card details to transaction

mrlumbu:
Pull out card details related data from event digest and set on transaction


jankowiakmaria:
Can be closed.
This https://github.com/alphagov/pay-ledger/pull/85 supersedes it.

PR #84, opened 2019-07-03
PP-5359 separate entity and model and implement v1 search

SandorArpa:
This PR separates the transaction model from transaction entity. The combined class was getting messy and overcrowded.
This PR also removes `EventDetailsDigest`. It changes `EventDigest`'s structure with the event payload (all the event data combined into a key-value pair map).  `EventMessageHandler` lets `TransactionService` deal with the transaction data (above map). `TransactionService` then creates both `TransactionEntity` and `ConvertedTransactionDetails`. Latter one is solely to convert some of the payload that will be stored in `transaction_details` jsonb field in the `transaction` table.
The PR also implements the remaining query parameters from `connector`'s `v1` search endpoint. (see https://payments-platform.atlassian.net/browse/PP-5359?oldIssueView=true).  Had to go old school when implementing `IN(xxx, xxx)` in `TransactionSearchParams`. JDBI3 fluent API does not bind List<String> params.
Also updated/removed/added tests accordingly.


SandorArpa:
wrong branch

PR #85, opened 2019-07-03
PP-5359 separate entity and model and implement v1 search

SandorArpa:
This PR separates the transaction model from transaction entity. The combined class was getting messy and overcrowded.
This PR also removes `EventDetailsDigest`. It changes `EventDigest`'s structure with the event payload (all the event data combined into a key-value pair map).  `EventMessageHandler` lets `TransactionService` deal with the transaction data (above map). `TransactionService` then creates both `TransactionEntity` and `ConvertedTransactionDetails`. Latter one is solely to convert some of the payload that will be stored in `transaction_details` jsonb field in the `transaction` table.
The PR also implements the remaining query parameters from `connector`'s `v1` search endpoint. (see https://payments-platform.atlassian.net/browse/PP-5359?oldIssueView=true).  Had to go old school when implementing `IN(xxx, xxx)` in `TransactionSearchParams`. JDBI3 fluent API does not bind List<String> params.
Also updated/removed/added tests accordingly.


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -65,27 +81,29 @@ private TransactionView decorateWithLinks(TransactionView transactionView,
...
-                eventDigest.getEventDetailsDigest().getReturnUrl(),
-                eventDigest.getEventDetailsDigest().getEmail(),
-                eventDigest.getEventDetailsDigest().getPaymentProvider(),
-                eventDigest.getMostRecentEventTimestamp(),
-                null,
-                eventDigest.getEventDetailsDigest().getDelayedCapture(),
-                null,
-                eventDigest.getEventCount()
-        );
+    public TransactionEntity convertToTransaction(EventDigest eventDigest) {
```
jankowiakmaria: Would be better to have it out of the service, but can be cleaned up later on.

src/test/java/uk/gov/pay/ledger/util/fixture/QueueEventFixture.java
```diff
@@ -50,6 +54,49 @@ public QueueEventFixture withEventData(String eventData) {
         return this;
     }
 
+    public QueueEventFixture withGatewayAccountId(String gatewayAccountId) {
+        this.gatewayAccountId = gatewayAccountId;
+        return this;
+    }
+
+    public QueueEventFixture withEventData(SalientEventType eventType) {
```
jankowiakmaria: It's not obvious what the method does - could be renamed.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -135,33 +141,29 @@ public String generateQuery() {
...
+            sb.append(" AND t.state = :" + STATE_FIELD);
         }
-
         if (refundStates != null) {
-            sb.append(" AND false");
+            //TODO implement
         }
         if (cardBrands != null && !cardBrands.isEmpty()) {
-            sb.append(" AND false");
+            sb.append(" AND t.card_brand IN(" + inClause(cardBrands) + ")");
```
jankowiakmaria: inClause generated manually (not by binding) is the worrying thing.

It seems it's theoretically possible to bind the list instead of manually generating string (http://jdbi.org/#_binding_arguments), but I'll have a look at it in separate PR (this one is already to big)

src/main/java/uk/gov/pay/ledger/event/model/SalientEventType.java
```diff
@@ -5,7 +5,7 @@
 
 public enum SalientEventType {
     PAYMENT_CREATED,
-    AUTHORISATION_SUCCESSFUL;
+    PAYMENT_DETAILS_EVENT;
```
jankowiakmaria: Not sure whether `PAYMENT_DETAILS_EVENT` is a salient event type.

PR #89, opened 2019-07-05
PP-5359: bind params in `IN` clause

jankowiakmaria:
* updated `TransactionSearchParams` and `TransactionDao` to bind the params instead of manually constructing string with values
* a bit of refactor in `TransactionDao` to remove duplicate code (probably can be done more of it)

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.ledger.transaction.dao;
 
 import com.google.inject.Inject;
+import org.jdbi.v3.core.Handle;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.statement.Query;
 import uk.gov.pay.ledger.transaction.dao.mapper.TransactionMapper;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/search/common/CommaDelimitedSetParameter.java
```diff
@@ -1,30 +1,19 @@
...
-    }
-
-    public Stream<String> stream() {
-        return elements.stream();
-    }
-
-    public boolean has(String element) {
-        return elements.contains(element);
+                ? new ArrayList<>()
+                : List.of(queryString.split(","));
```
sfount: Nice 👍

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -71,27 +72,36 @@ public TransactionDao(Jdbi jdbi) {
...
             return query
                     .mapTo(Long.class)
                     .findOnly();
         });
     }
 
+    private Query getQuery(TransactionSearchParams searchParams, Handle handle, String searchCountQueryString) {
+        String searchExtraFields = searchParams.generateQuery();
+        Query query = handle.createQuery(searchCountQueryString.replace(":searchExtraFields", searchExtraFields));
+        searchParams.getQueryMap().forEach((k, v) -> {
```
sfount: This took me a minute to wrap my head around!

Even though it's more verbose maybe something like `searchKey` and `searchValue` would help with understanding what this is doing. 

Or moving it to a static named method that's passed in to `forEach`, this is probably simple enough to not need it.

jankowiakmaria: I see. Thanks. Will have a look at it separately.

PR #90, opened 2019-07-05
PP-5359: clean up

jankowiakmaria:
* moved creating TransactionEntity to its own factory (from Transaction Service) <- only moved it, haven't changed the way it's been done
* test for ^^
* renamed fixture method

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
@@ -1,13 +1,12 @@
 package uk.gov.pay.ledger.transaction.service;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
```diff
@@ -43,8 +39,8 @@
     private TransactionSearchParams searchParams;
 
     @Before
-    public void setUp() throws MalformedURLException {
-        transactionService = new TransactionService(mockTransactionDao);
+    public void setUp() {
+        transactionService = new TransactionService(mockTransactionDao, new TransactionEntityFactory(new ObjectMapper()));
```
heathd: I don't think we use/rely on any of the dropwizard customisations but I *think* it's recommended to use `Jackson.newObjectMapper()` which builds an object mapper with some extra plugins, or at least that's a more realistic simulation of what would get provided through DI

jankowiakmaria: I see. Thanks!

PR #108, opened 2019-07-15
PP-5359: handle null fields for search

jankowiakmaria:
* added static factory methods to Address and CardDetails to handle lack of data
* added JsonInclude annotation to TransactionView to show only non-null properties


jankowiakmaria:
Let me know if there is a better way to do it.

PR #109, opened 2019-07-15
PP-5360: additional fields in transaction table

jankowiakmaria:
* added fields related to refund and settlement summary
* added fields related to fees (net_amount and total_amount)

### DIFF
--- a/src/main/resources/migrations/00012_add_fields_to_transaction_table.sql
+++ b/src/main/resources/migrations/00012_add_fields_to_transaction_table.sql
@@ -1,0 +1,12 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_fields_to_transaction_table
+
+ALTER TABLE transaction
+    ADD COLUMN net_amount BIGINT,
+    ADD COLUMN total_amount BIGINT,
+    ADD COLUMN settlement_submitted_time TIMESTAMP WITH TIME ZONE,
+    ADD COLUMN settled_time TIMESTAMP WITH TIME ZONE,
+    ADD COLUMN refund_status VARCHAR(100),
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00012_add_fields_to_transaction_table.sql
```diff
@@ -0,0 +1,12 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_fields_to_transaction_table
+
+ALTER TABLE transaction
+    ADD COLUMN net_amount BIGINT,
+    ADD COLUMN total_amount BIGINT,
+    ADD COLUMN settlement_submitted_time TIMESTAMP WITH TIME ZONE,
+    ADD COLUMN settled_time TIMESTAMP WITH TIME ZONE,
+    ADD COLUMN refund_status TEXT,
```
heathd: (not a blocker)

would there be benefit of using `varchar(N)` .. where `N` is say `100`? I know there's no performance advantage of varchar over text, but perhaps making it explicit that the size of data that can be stored has an upper limit (text is 'unlimited')

jankowiakmaria: Possibly - I don't mind.
I just wanted to make it consistent (type-wise) with `status` https://github.com/alphagov/pay-ledger/blob/32b426629446ca325fa7e6f73c77a1cda7e30114/src/main/resources/migrations/00001_create_table_transaction.sql#L12

jankowiakmaria: Updated to `VARCHAR(100)`. Need reapproval.

PR #114, opened 2019-07-17
PP-5388: salient event external state map

sfount:
* add missing external states
* add latest version of events and their map to states

Depends on https://github.com/alphagov/pay-connector/pull/1503


src/main/java/uk/gov/pay/ledger/event/model/SalientEventType.java
```diff
@@ -5,7 +5,32 @@
 
 public enum SalientEventType {
```
mrlumbu: I wonder if we should rename this to `EventType` - feel like the salient concept has died

src/main/java/uk/gov/pay/ledger/transaction/state/TransactionState.java
```diff
@@ -46,10 +54,34 @@ public String getMessage() {
         return message;
     }
 
-    private static final Map<SalientEventType, TransactionState> EVENT_TYPE_TRANSACTION_STATE_MAP = Map.of(
-            SalientEventType.PAYMENT_CREATED, CREATED,
-            SalientEventType.AUTHORISATION_SUCCESSFUL, SUBMITTED
-    );
+    private static final Map<SalientEventType, TransactionState> EVENT_TYPE_TRANSACTION_STATE_MAP =
+            ImmutableMap.<SalientEventType, TransactionState>builder()
```
mrlumbu: `Map.of` returns an immutable map (`ImmutableMap` was a google util for pre Java 11 times)

src/main/java/uk/gov/pay/ledger/event/model/SalientEventType.java
sfount: @heathd suggested just using Strings in this map as the contract with Connector events is based on their name and the object isn't used anywhere else, what do you think? 

Happy to rename to `EventType` in the interim

src/main/java/uk/gov/pay/ledger/transaction/state/TransactionState.java
sfount: `Map.of` has a size limit for the number of pairs. As you've suggested I'll use `Map.ofEntries` in favour of an external library

sfount:
@mrlumbu 
https://github.com/alphagov/pay-ledger/pull/114/commits/df5b434eebfb3e13f25984788dd56d4bb589d82b
* `SalientEventType` -> `EventType` 
* `ImmutableMap` -> `Map.ofEntries`

PR #115, opened 2019-07-18
BAU: clean up TransactionEntity

jankowiakmaria:
* introduce lambok dependency
* use builder from lambok
* update usages of TransactionEntity (creating new instance)
* replace getters and setters with lambok annotations


jankowiakmaria:
Closing - based on previous conversations and decision not to use it.

PR #116, opened 2019-07-18
BAU: add TransactionEntity builder

jankowiakmaria:
Instead of https://github.com/alphagov/pay-ledger/pull/115
* added builder to TransactionEntity class
* cleaned up usages (new instance creation) of TransactionEntity


jankowiakmaria:
Naming convention is not established in Ledger.
In Connector it's mostly used with `with`.
In PublicAPO it't used either with `with` or just field name as a name of the setting function.

PR #117, opened 2019-07-18
BAU fix TransactionState.from method

SandorArpa:
## WHAT
- optional orElseGet(null) throws a NullPointerException. Switched to orElse(null)


SandorArpa:
https://github.com/alphagov/pay-ledger/pull/118 superseeds it

PR #110, opened 2019-07-16
PP-5360 get one transaction

jankowiakmaria:
depends on https://github.com/alphagov/pay-publicapi/pull/580

PR #120, opened 2019-07-18
BAU: updated signature of getZonedDateTime

jankowiakmaria:
* returns Optional
* behaviour of the non-existing date is defined in "consumer"


sfount:
Addresses comments in https://github.com/alphagov/pay-ledger/pull/110

PR #126, opened 2019-07-19
PP-5460: new columns in transaction table

jankowiakmaria:
* added `fee` ~and `corporate_surcharge`~ column~s~ to Transaction table - migration

### DIFF
--- a/src/main/resources/migrations/00013_add_fee_columns_to_transaction_table.sql
+++ b/src/main/resources/migrations/00013_add_fee_columns_to_transaction_table.sql
@@ -1,7 +1,0 @@
---liquibase formatted sql
-
---changeset uk.gov.pay:add_fee_columns_to_transaction_table
-
-ALTER TABLE transaction
-    ADD COLUMN fee BIGINT,
-    ADD COLUMN corporate_surcharge BIGINT;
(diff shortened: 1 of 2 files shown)


src/main/resources/migrations/00013_add_fee_columns_to_transaction_table.sql
```diff
@@ -0,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_fee_columns_to_transaction_table
+
+ALTER TABLE transaction
+    ADD COLUMN fee BIGINT,
+    ADD COLUMN corporate_surcharge BIGINT;
```
sfount: I think `corporate_surcharge` is currently stored in the `transaction_details` column on transactions. Do we need it at the top level too?

jankowiakmaria: No we don't need to - it would be one or another.
The story describes appropriate columns being available, and after writing the story there was a discussion to do it differently.

jankowiakmaria: Will remove `corporate_surcharge` from top level.

PR #127, opened 2019-07-19
PP-5460: populate fields

jankowiakmaria:
populate `net_amount` and `total_amount`
rename factory method (to follow convention)


src/main/java/uk/gov/pay/ledger/event/model/TransactionEntityFactory.java
```diff
@@ -21,7 +21,7 @@ public TransactionEntityFactory(ObjectMapper objectMapper){
         this.objectMapper = objectMapper;
     }
 
-    public TransactionEntity from(EventDigest eventDigest) {
```
sfount: Nice 👍

PR #128, opened 2019-07-19
PP-5457 Add columns to support refunds

kbottla:
## WHAT
- Added columns `gateway_transaction_id`, `origin_transaction_id`, `transaction_type` to support refund related endpoints
- Transaction type is currently set to nullable.
    Todo:
       - Update code to always populate transaction_type
       - Update transactoion_type to not null

Co-authored-by: @sfount

### DIFF
--- a/src/main/resources/migrations/00014_add_fields_for_refunds.sql
+++ b/src/main/resources/migrations/00014_add_fields_for_refunds.sql
@@ -1,0 +1,12 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_refund_related_fields_to_transaction_table
+
+CREATE type transaction_type as enum ('PAYMENT', 'REFUND');
+
+ALTER TABLE transaction
+    ADD COLUMN gateway_transaction_id text,
+    ADD COLUMN type transaction_type ,
+    ADD COLUMN origin_transaction_id VARCHAR(26) references transaction(external_id);
+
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00013_add_fields_for_refunds.sql
```diff
@@ -0,0 +1,12 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_refund_related_fields_to_transaction_table
+
+CREATE type transaction_type_t as enum ('PAYMENT', 'REFUND');
```
jankowiakmaria: do you name types with suffix `_t`? Is it needed?

kbottla: - Updated to `00014..`
- We don't have precedent for types yet. Similar to `idx` we have added `_t` for type

jankowiakmaria: I just wonder whether it makes sense to suffix types. Especially that the name itself mentions `type`.
`idx` is more descriptive than `t` (which could even mean table, right?)
Feels a bit like Hungarian notation. That's why I wonder whether it's a good practice.
I can approve, but I still think it should be reconsidered.

sfount: The `type` in the name is not to do with an type - it's the transaction *type* type - if that makes sense? transaction_type exists outside the fact that this is the type for that column

PR #129, opened 2019-07-19
PP-5460 populate corporate_surcharge in transactionDetails

jankowiakmaria:
* add corporate_surcharge to ConvertedTransactionDetails to make sure it's part of the transaction_details column of Transaction table


jankowiakmaria:
The contract test is failing because the connector hasn't been yet updated to publish this property.

PR #140, opened 2019-07-22
PP-5457 Make type (transaction table) to not nullable field

kbottla:


src/main/resources/migrations/00015_make_transaction_type_required
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:make_type_as_required_field_in_transaction_table
+
+UPDATE transaction set type = 'PAYMENT' where type is not null;
```
sfount: Should this be `where type is null?`Maybe there's a GitHub glitch as I thought this was changed

PR #146, opened 2019-07-23
BAU: clean up transaction_details

jankowiakmaria:
* put json properties in transaction_details only if they have value (JSON annotation)
* chaged `as<type>` to `<type>Value` to retrieve value insted of making conversions (null won't be treated as "null")
* updated tests to reflect changes in the behaviour

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/PaymentFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/PaymentFactory.java
@@ -53,8 +53,8 @@ public class PaymentFactory {
                     TransactionState.from(entity.getState()), safeGetAsString(transactionDetails, "language"),
                     entity.getExternalId(), safeGetAsString(transactionDetails, "return_url"), entity.getEmail(),
                     safeGetAsString(transactionDetails, "payment_provider"), entity.getCreatedDate(),
-                    cardDetails, Boolean.valueOf(safeGetAsString(transactionDetails, "delayed_capture")),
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/PaymentFactory.java
```diff
@@ -53,8 +53,8 @@ public Payment createTransactionEntity(TransactionEntity entity) {
                     TransactionState.from(entity.getState()), safeGetAsString(transactionDetails, "language"),
                     entity.getExternalId(), safeGetAsString(transactionDetails, "return_url"), entity.getEmail(),
                     safeGetAsString(transactionDetails, "payment_provider"), entity.getCreatedDate(),
-                    cardDetails, Boolean.valueOf(safeGetAsString(transactionDetails, "delayed_capture")),
-                    metadata, entity.getEventCount(), safeGetAsString(transactionDetails, "gateway_transaction_id"),
+                    cardDetails, safeGetAsBoolean(transactionDetails, "delayed_capture", false), metadata,
```
sfount: Just want to double check that a `false` value is the same as nothing being set. Should this be the default value or does it not being set (by the service?) have meaning.

jankowiakmaria: This change duplicates the previous behaviour (if null or false -> false).

I've checked in Connector and default value of `delayedCapture` is false when creating new charge.

PR #150, opened 2019-07-23
PP-5403 Integration test for refund event

kbottla:


src/test/java/uk/gov/pay/ledger/queue/QueueMessageReceiverIT.java
```diff
@@ -84,4 +84,34 @@ public void shouldContinueToHandleMessagesFromQueueForDownstreamExceptions() thr
...
+                .insert(rule.getSqsClient());
+
+        Thread.sleep(500);
+
+        given().port(rule.getAppRule().getLocalPort())
+                .contentType(JSON)
+                .get("/v1/transaction/" + resourceExternalId)
+                .then()
+                .statusCode(200)
+                .body("charge_id", is(resourceExternalId))
```
sfount: Why does the refund have a `charge_id`? If this isn't expected behaviour potentially there should be a comment that reminds us to update this once the read end points have been updated.

kbottla: We don't have separate endpoint to search for refunds. Added a comment to use refunds endpoint when available.

sfount: Great - I'm doing to add a ticket just to make sure!

PR #154, opened 2019-07-24
Fix - for missing event type and mapping

kbottla:
## WHAT 

- `USER_APPROVED_FOR_CAPTURE` : Event is missing transaction state which is resulting in errors
     `Error during handling the event message. [id=xxxxxx] [queueMessageId=xxxxx] [errorMessage=null]`
- `PAYMENT_DETAILS_ENTERED` : Missing from event types. Added event type and transaction state mapping


src/main/java/uk/gov/pay/ledger/transaction/state/TransactionState.java
```diff
@@ -62,6 +62,7 @@ public String getMessage() {
                     Map.entry(EventType.PAYMENT_CREATED, CREATED),
                     Map.entry(EventType.PAYMENT_STARTED, STARTED),
                     Map.entry(EventType.PAYMENT_EXPIRED, FAILED_EXPIRED),
+                    Map.entry(EventType.PAYMENT_DETAILS_ENTERED, STARTED),
```
sfount: `sudo rm -rf`

sfount: Reference https://github.com/alphagov/pay-ledger/pull/157

PR #157, opened 2019-07-24
PP-5403 Support non state transition events

sfount:
* use the most recent `s/salient/state-transitioning/g` event to determine event digest state

Implicitly this sets the standard that: 
* ledger shouldn't need to know about events that don't transition state (directly map to external states)
* any number of payload events can be related to a given digest, as long as at least one event maps to an external state ledger knows about

Question: should it be enforced that registered enum `EventType` has a map to an external state in `TransactionState`

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
+++ b/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
@@ -1,6 +1,7 @@
 package uk.gov.pay.ledger.event.model;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
+import uk.gov.pay.ledger.transaction.state.TransactionState;
 
 import java.io.IOException;
 import java.time.ZonedDateTime;
@@ -16,6 +17,7 @@ public class EventDigest {
     private final String resourceExternalId;
     private final String parentResourceExternalId;
     private final EventType mostRecentEventType;
+    private final String mostRecentEventState;
     private Integer eventCount;
     private Map<String, Object> eventPayload;
     private final ZonedDateTime eventCreatedDate;
@@ -26,6 +28,7 @@ public class EventDigest {
             ResourceType resourceType,
             String resourceExternalId,
             String parentResourceExternalId,
+            String mostRecentEventState,
             Integer eventCount,
             Map<String, Object> eventPayload,
             ZonedDateTime eventCreatedDate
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
```diff
@@ -63,6 +68,7 @@ public static EventDigest fromEventList(List<Event> events) {
                 latestEvent.getResourceType(),
                 latestEvent.getResourceExternalId(),
                 latestEvent.getParentResourceExternalId(),
+                TransactionState.fromEventType(latestSalientEventType).getState(),
```
mrlumbu: I thought it might be nice if the event digest reports on the event side of things (i.e. latest event) and the transaction side of things does the conversion to `TransactionState` - this is how it was before. Thoughts?

sfount: Yeah that separation makes sense, will update

src/test/java/uk/gov/pay/ledger/event/model/TransactionEntityFactoryTest.java
```diff
@@ -103,4 +108,34 @@ public void fromShouldConvertEventDigestToTransactionForChildResource() {
...
+                .withResourceType(ResourceType.PAYMENT)
+                .toEntity();
+
+        thrown.expect(RuntimeException.class);
+        thrown.expectMessage("No supported external state transition events found for digest");
+        EventDigest.fromEventList(List.of(firstNonSalientEvent, secondNonSalientEvent));
+    }
+
+    @Test
+    public void fromShouldGetLatestEventDigestState() {
```
mrlumbu: not sure if I understand this test name

src/test/java/uk/gov/pay/ledger/event/model/TransactionEntityFactoryTest.java
```diff
@@ -103,4 +108,34 @@ public void fromShouldConvertEventDigestToTransactionForChildResource() {
         assertThat(transactionEntity.getParentExternalId(), is(parentResourceExternalId));
         assertThat(transactionEntity.getState(), is("submitted"));
     }
+
+    @Test
+    public void fromShouldRejectEventDigestWithoutStateTransitionEvents() throws Exception {
```
mrlumbu: are we calling them `salient` or `statetransition` events

sfount: have updated everything back to `Team Salient`

PR #158, opened 2019-07-24
PP-5403 Process additional refund event field

kbottla:
## WHAT 

Refund submitted event includes `refunded_by` field in the payload which is not directly mapped to Transaction table. Updated to map `refunded_by` to `transaction_details` json blob


src/main/java/uk/gov/pay/ledger/transaction/search/model/ConvertedTransactionDetails.java
```diff
@@ -37,5 +38,7 @@ public ConvertedTransactionDetails() {}
     @JsonProperty("gateway_transaction_id")
     private String gatewayTransactionId;
     @JsonProperty("corporate_surcharge")
-    private Long corporate_surcharge;
+    private Long corporateSurcharge;
+    @JsonProperty("refunded_by")
```
sfount: This name is much better 🎉

PR #170, opened 2019-07-29
PP-5460: pact test for PaymentDetailsEntered event

jankowiakmaria:


src/test/java/uk/gov/pay/ledger/pact/PaymentDetailsEnteredEventQueueContractTest.java
```diff
@@ -0,0 +1,94 @@
...
+        EventDao eventDao = appRule.getJdbi().onDemand(EventDao.class);
+
+        await().atMost(1, TimeUnit.SECONDS).until(
+                () -> transactionDao.findTransactionByExternalId(externalId).isPresent()
+        );
+
+        Optional<TransactionEntity> transaction = transactionDao.findTransactionByExternalId(externalId);
+        assertThat(transaction.isPresent(), is(true));
+        assertThat(transaction.get().getExternalId(), is(externalId));
+        assertThat(transaction.get().getEmail(), is("j.doe@example.org"));
```
kbottla: Could get away with hard coded string if fixture is used as here https://github.com/alphagov/pay-ledger/blob/55cf5db6d546d996da2aeeb76d9ae5158ee6963d/src/test/java/uk/gov/pay/ledger/pact/RefundCreatedByUserEventQueueContractTest.java#L41

src/test/java/uk/gov/pay/ledger/pact/PaymentDetailsEnteredEventQueueContractTest.java
```diff
@@ -0,0 +1,94 @@
...
+        assertThat(transaction.get().getCardBrand(), is("visa"));
+        assertThat(transaction.get().getFirstDigitsCardNumber(), is("424242"));
+        assertThat(transaction.get().getLastDigitsCardNumber(), is("4242"));
+        assertThat(transaction.get().getCardholderName(), is("J citizen"));
+        assertThat(transaction.get().getTransactionDetails(), containsString("\"language\": \"en\""));
+        assertThat(transaction.get().getTransactionDetails(), containsString("\"corporate_surcharge\": 5"));
```
kbottla: Similarly we can use fixture for transacationDetails too https://github.com/alphagov/pay-ledger/blob/55cf5db6d546d996da2aeeb76d9ae5158ee6963d/src/test/java/uk/gov/pay/ledger/pact/RefundCreatedByUserEventQueueContractTest.java#L80

src/test/java/uk/gov/pay/ledger/pact/PaymentDetailsEnteredEventQueueContractTest.java
jankowiakmaria: I'm rather a huge fan of hardcoded strings in tests.

src/test/java/uk/gov/pay/ledger/pact/PaymentDetailsEnteredEventQueueContractTest.java
jankowiakmaria: I'm rather a huge fan of hardcoded strings in tests.

src/test/java/uk/gov/pay/ledger/pact/PaymentDetailsEnteredEventQueueContractTest.java
kbottla: Not a big issue but we are ending up with different patterns with in same code base which we might want to avoid

PR #173, opened 2019-07-30
PP-5454 Add DAO class for transaction events

kbottla:
## WHAT 

- Adds DAO class `TransactionEventDao` and method to query events by external_id & gateway_account_id
- Uses both Transaction and Event tables as `gateway_account_id` & `amount` (should be available for each event) are available in Transaction and the rest of information in Event.
- Filters events by `parent_external_id` to find refunds and related events (avoids self joining transaction table)
- Relevant integration tests


kbottla:
Superseded by https://github.com/alphagov/pay-ledger/pull/177

PR #183, opened 2019-07-31
PP-5463 Snake case for transaction entity

mrlumbu:
All fields in TransactionEntity should be mapped to snake case
when serialising to/deserialising from JSON


jankowiakmaria:
Could you rebase, please?

PR #188, opened 2019-08-01
PP-5419 refund class

jankowiakmaria:
The last thing is (outside of the scope of this PR):  getting rid of links - they give no value and now they're misleading (refund will get exactly the same links as the regular payment)


src/main/java/uk/gov/pay/ledger/transaction/search/model/TransactionView.java
```diff
@@ -45,6 +47,7 @@
     private RefundSummary refundSummary;
     private SettlementSummary settlementSummary;
     private Map<String, Object> metadata;
+    private String refundedBy;
     private List<Link> links = new ArrayList<>();
 
     public TransactionView(Builder builder) {
```
mrlumbu: not really relevant to your changes, but this `TransactionView` should really just be a wrapper round `Transaction` and proxy calls through, rather than having to set every field again. And that would mean you wouldn't have to care about the sub-type

jankowiakmaria: Yeah, I guess it would be good to clean it up at some point.

PR #194, opened 2019-08-02
PP-5491 Correct transaction state handling

mrlumbu:
Previously we were doing the following:

### Processing an Incoming event
- Make event digest
- Calculate TransactionState from latest salient event
- Store transaction state as Transaction.getStatus()

### Getting a transaction
- Get transaction status from db
- Convert to first TransactionState with matching status

This could lead to the following happening:

- Latest event is CANCEL_BY_USER_FAILED
- Therefore TransactionState is FAILED_CANCELLED
- Transaction state stored in DB as cancelled
- Transaction retrieved
- TransactionState calculated as CANCELLED

i.e. we could lose information about the real TransactionState and
therefore report the wrong state.

Now we store the actual name of the transaction state in the DB,
so we do not lose any information. I have had to adapt the search
so that now if someone searches for cancelled, we actual go and retrieve
transactions with state FAILED_CANCELLED and CANCELLED

This means all existing transaction states in DB will be wrong. This is fine as we will truncate existing data

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
@@ -3,13 +3,16 @@ package uk.gov.pay.ledger.transaction.search.common;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.ledger.transaction.model.TransactionType;
+import uk.gov.pay.ledger.transaction.state.TransactionState;
 
 import javax.ws.rs.QueryParam;
 import java.time.ZonedDateTime;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -150,7 +153,7 @@ public String generateQuery() {
             //TODO implement
         }
         if (isNotBlank(state)) {
-            sb.append(" AND t.state = :" + STATE_FIELD);
+            sb.append(" AND t.state in  (<" + STATE_FIELD + ">)");
```
jankowiakmaria: [tiny]
```suggestion
            sb.append(" AND t.state IN (<" + STATE_FIELD + ">)");
```

PR #199, opened 2019-08-05
PP-5462 Store and display settlement summary

mrlumbu:
Use capture_submitted_time consistently throughout date. This is an attempt to be consistent - we have used _date everywhere else for timestamps, and this lines up with captured_date
This commit ensures capture_submitted_date  persisted and retrieved correctly, using a pact test to test the end to end behaviour.

_edit_: I have now also done the following, based on @kbottla comment and discussion:
_I have changed how transaction_details are stored. Instead of mapping to an object (which means we have to specify each field we wish to remain in transaction details) we now just write the whole thing into the db. This has the advantage of not having to update ledger each time a field that we do not consider to be a top level column is added to an event somewhere._


src/main/java/uk/gov/pay/ledger/app/LedgerModule.java
```diff
@@ -35,6 +36,15 @@ protected void configure() {
...
     }
 
+    @Provides
+    @Singleton
+    public ObjectMapper provideObjectMapper() {
+        ObjectMapper objectMapper = environment.getObjectMapper();
+        objectMapper.findAndRegisterModules();
+
+        return objectMapper;
+    }
```
mrlumbu: this gets Jackson to serialise/deserialise dates automatically

src/main/resources/migrations/00018_rename_settlement_coumns.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE transaction rename COLUMN settlement_submitted_time to capture_submitted_date;
```
kbottla: Can we have this migration as separate PR or this might be opportunity to move it to `transaction_details` if it is not required at table level ?

mrlumbu: yeah, maybe we should move to details - I'll do that

src/main/java/uk/gov/pay/ledger/event/model/TransactionEntityFactory.java
```diff
@@ -41,9 +40,8 @@ public TransactionEntity create(EventDigest eventDigest) {
     }
 
     private String convertToTransactionDetails(Map<String, Object> transactionPayload) {
-        ConvertedTransactionDetails details = objectMapper.convertValue(transactionPayload, ConvertedTransactionDetails.class);
         try {
-            return objectMapper.writeValueAsString(details);
+            return objectMapper.writeValueAsString(transactionPayload);
```
kbottla: We may be saving top level details into `trasactionDetails` too. Couple of issues but may not be a blocker
- redundant data as fields (email,card digits.. ) are at top level and also in transactionDetails
- Might add confusion in future and may end up using top level at some places and `transactionDetails` JSON fields at other places.

May be we can diff top level columns data out of `transactionDetails` ?

mrlumbu: We discussed this on Tuesday (I thought you had been involved in the discussion but evidently not). As I said in my comment, the reason is:
_I have changed how transaction_details are stored. Instead of mapping to an object (which means we have to specify each field we wish to remain in transaction details) we now just write the whole thing into the db. This has the advantage of not having to update ledger each time a field that we do not consider to be a top level column is added to an event somewhere._
Personally I don't think there's anything confusing about having the data in details as well, and the diff option could be fiddly if there are different field names in the event versus in the transaction etc.

PR #208, opened 2019-08-08
PP-5541 Use refunded amount column in favour of refund submitted

sfount:
(Waiting on pacts given services master builds failing)

* database migration for column name change
* column name matches event structure defined by https://github.com/alphagov/pay-connector/pull/1515
* follow up PR to update payment resource model

Requires initial PR to introduce the column (through migration), followed up by a cleanup removing the extra column.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/model/RefundSummary.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/model/RefundSummary.java
@@ -17,23 +17,26 @@ public class RefundSummary {
 
     private Long amountSubmitted;
 
+    private Long amountRefunded;
+
     public RefundSummary() {
     }
 
-    public RefundSummary(String status, Long amountAvailable, Long amountSubmitted) {
+    public RefundSummary(String status, Long amountAvailable, Long amountRefunded) {
         this.status = status;
         this.amountAvailable = amountAvailable;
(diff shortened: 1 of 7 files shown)


sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/211

src/main/java/uk/gov/pay/ledger/transaction/search/model/RefundSummary.java
```diff
@@ -17,23 +17,26 @@
...
     public RefundSummary() {
     }
 
-    public RefundSummary(String status, Long amountAvailable, Long amountSubmitted) {
+    public RefundSummary(String status, Long amountAvailable, Long amountRefunded) {
         this.status = status;
         this.amountAvailable = amountAvailable;
-        this.amountSubmitted = amountSubmitted;
+        this.amountRefunded = amountRefunded;
+        this.amountSubmitted = amountRefunded;
```
jankowiakmaria: what's the plan for `amountSubmitted`? Right now there are two fields in this class that are populated with `amountRefunded`.

sfount: It should be removed with the old database column, we were somewhat restricted by failing master builds stopping us from updating pact tests

PR #222, opened 2019-08-12
PP-5527: update TransactionMapper

jankowiakmaria:
* in order to be able to read null value from db (instead of defaulting to the value of long primitive type)

(extended to other Long values not only `fee`)


src/main/java/uk/gov/pay/ledger/transaction/dao/mapper/TransactionMapper.java
```diff
@@ -33,18 +33,23 @@ public TransactionEntity map(ResultSet rs, StatementContext ctx) throws SQLExcep
...
-                .withRefundAmountAvailable(rs.getLong("refund_amount_available"))
-                .withFee(rs.getLong("fee"))
+                .withRefundAmountRefunded(getLongWithNullCheck(rs, "refund_amount_refunded"))
+                .withRefundAmountAvailable(getLongWithNullCheck(rs, "refund_amount_available"))
+                .withFee(getLongWithNullCheck(rs, "fee"))
                 .withTransactionType(rs.getString("type"))
                 .build();
     }
 
+    private Long getLongWithNullCheck(ResultSet rs, String columnName) throws SQLException {
```
sfount: 🍠

PR #228, opened 2019-08-13
PP-5498: add query parameters to get transaction endpoint

jankowiakmaria:
* add `transaction_type` and `parent_external_id` optional query parameters
* extended transaction Dao method
* updated (and added new) integration test

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
@@ -6,6 +6,7 @@ import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.statement.Query;
 import uk.gov.pay.ledger.transaction.dao.mapper.TransactionMapper;
 import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
+import uk.gov.pay.ledger.transaction.model.TransactionType;
 import uk.gov.pay.ledger.transaction.search.common.TransactionSearchParams;
 
 import java.util.List;
@@ -20,7 +21,9 @@ public class TransactionDao {
 
     private static final String FIND_TRANSACTION_BY_EXTERNAL_ID_AND_GATEWAY_ACCOUNT_ID = "SELECT * FROM transaction " +
             "WHERE external_id = :externalId " +
-            "and gateway_account_id = :gatewayAccountId";
+            "AND gateway_account_id = :gatewayAccountId " +
+            "AND (:transactionType::transaction_type is NULL OR type = :transactionType::transaction_type) " +
+            "AND (:parentExternalId is NULL OR parent_external_id = :parentExternalId)";
 
     private static final String FIND_TRANSACTIONS_BY_EXTERNAL_OR_PARENT_ID_AND_GATEWAY_ACCOUNT_ID = "SELECT * FROM transaction " +
             "WHERE (external_id = :externalId or parent_external_id = :externalId) " +
@@ -120,11 +123,13 @@ public class TransactionDao {
         this.jdbi = jdbi;
     }
 
-    public Optional<TransactionEntity> findTransactionByExternalIdAndGatewayAccountId(String externalId, String gatewayAccountId) {
+    public Optional<TransactionEntity> findTransaction(String externalId, String gatewayAccountId, TransactionType transactionType, String parentTransactionExternalId) {
         return jdbi.withHandle(handle ->
                 handle.createQuery(FIND_TRANSACTION_BY_EXTERNAL_ID_AND_GATEWAY_ACCOUNT_ID)
                         .bind("externalId", externalId)
                         .bind("gatewayAccountId", gatewayAccountId)
+                        .bind("transactionType", transactionType)
+                        .bind("parentExternalId", parentTransactionExternalId)
                         .map(new TransactionMapper())
                         .findFirst());
     }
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -20,7 +21,9 @@
 
     private static final String FIND_TRANSACTION_BY_EXTERNAL_ID_AND_GATEWAY_ACCOUNT_ID = "SELECT * FROM transaction " +
             "WHERE external_id = :externalId " +
-            "and gateway_account_id = :gatewayAccountId";
+            "AND gateway_account_id = :gatewayAccountId " +
+            "AND (:transactionType::transaction_type is NULL OR type = :transactionType::transaction_type) " +
```
kbottla: Makes me worried why this change didn't break existing pacts. For `a transaction with created state exist` we don't currently pass transactionType parameter and as the query should now be `transaction_type is NULL` (which is not currently) I am expecting pact test to break unless I am missing something ?

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -20,7 +21,9 @@
...
+            "AND (:parentExternalId is NULL OR parent_external_id = :parentExternalId)";
```
kbottla: Are we making it mandatory to provide `parentExternalId` for get one refund transaction ?

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -120,11 +123,13 @@ public TransactionDao(Jdbi jdbi) {
         this.jdbi = jdbi;
     }
 
-    public Optional<TransactionEntity> findTransactionByExternalIdAndGatewayAccountId(String externalId, String gatewayAccountId) {
+    public Optional<TransactionEntity> findTransactionByExternalIdAndGatewayAccountId(String externalId, String gatewayAccountId, TransactionType transactionType, String parentTransactionExternalId) {
```
kbottla: Method name could change as it's purpose for transactions with more filters

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
kbottla: Looks like we don't have a pact test for `get-one-transaction`

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
jankowiakmaria: I don't think it needs to be mandatory on ledger side. If it's not valid, it won't return any results.

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
jankowiakmaria: yeah, wasn't sure how to name it.

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
jankowiakmaria: `:transactionType::transaction_type is NULL` checks whether the passed parameter is null, not the value of the column in db.
For Payment we don't currently pass value for transaction type when we retrieve it in public api from ledger. (It will be done separately)

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
kbottla: This could break if `:parentExternalId` is an empty string which could be ignored ?

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
kbottla: I prefer constructing query if value is available. Not a great fan of using `NULL is NULL` condition but fine with it too.

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
kbottla: May be 
```suggestion
    public Optional<TransactionEntity> findTransaction(String externalId, String gatewayAccountId, TransactionType transactionType, String parentTransactionExternalId) {
```

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
jankowiakmaria: Should it be an empty string?

kbottla: This is the usual pattern for non-mandatory fields until it is different in this case 

https://github.com/alphagov/pay-ledger/blob/bc62be837a511359e961dcad08293e4c6947a943/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java#L271

jankowiakmaria: I don't think it's required in this case. Public api doesn't specify this parameter at all when calling ledger, so it shouldn't be an issue.
If we discover otherwise we can update the code to handle the scenario.

PR #234, opened 2019-08-14
PP-5555 Implement search for {payment,refund}_states

mrlumbu:
Bit fiddly because we have one set of states but the search is on
two different transaction types.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
@@ -8,11 +8,12 @@ import uk.gov.pay.ledger.transaction.state.TransactionState;
 import javax.ws.rs.QueryParam;
 import java.time.ZonedDateTime;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static java.lang.String.format;
-import static java.lang.String.join;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class TransactionSearchParams {
@@ -157,14 +158,11 @@ public class TransactionSearchParams {
         if (toDate != null) {
             sb.append(" AND t.created_date < :" + TO_DATE_FIELD);
         }
-        if (paymentStates != null && !paymentStates.isEmpty()) {
-            //TODO implement
+        if (isSet(paymentStates) || isSet(refundStates)) {
+            sb.append(" AND " + createStateFilter());
         }
         if (isNotBlank(state)) {
             sb.append(" AND t.state IN (<" + STATE_FIELD + ">)");
-        }
-        if (refundStates != null) {
-            //TODO implement
         }
         if (cardBrands != null && !cardBrands.isEmpty()) {
             sb.append(" AND t.card_brand IN (<" + CARD_BRAND_FIELD + ">)");
@@ -211,6 +209,12 @@ public class TransactionSearchParams {
             }
             if (cardBrands != null && !cardBrands.isEmpty()) {
                 queryMap.put(CARD_BRAND_FIELD, cardBrands.getParameters());
+            }
+            if (isSet(paymentStates)) {
+                queryMap.put(PAYMENT_STATES_FIELD, getTransactionStatesForStatuses(paymentStates));
+            }
+            if (isSet(refundStates)) {
+                queryMap.put(REFUND_STATES_FIELD, getTransactionStatesForStatuses(refundStates));
             }
             if (isNotBlank(lastDigitsCardNumber)) {
                 queryMap.put(LAST_DIGITS_CARD_NUMBER_FIELD, lastDigitsCardNumber);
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
```diff
@@ -187,6 +187,148 @@ public void shouldSearchUsingAllFieldsAndReturnAllFieldsCorrectly() {
...
+                        "&page=1" +
+                        "&display_size=5" +
+                        "&payment_states=created,submitted"
+                )
+                .then()
+                .statusCode(Response.Status.OK.getStatusCode())
+                .contentType(JSON)
+                .body("results[0].transaction_id", is(createdTransaction.getExternalId()))
+                .body("results[1].transaction_id", is(submittedTransaction.getExternalId()))
+                .body("results[1].transaction_id", is(nullValue()));
```
jankowiakmaria: Shouldn't this line be:
```suggestion
                .body("results[2].transaction_id", is(nullValue()));
```
?

That's what's breaking the build.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
```diff
@@ -187,6 +187,148 @@ public void shouldSearchUsingAllFieldsAndReturnAllFieldsCorrectly() {
...
+                        "&payment_states=submitted"
+                )
+                .then()
+                .statusCode(Response.Status.OK.getStatusCode())
+                .contentType(JSON)
+                .body("results[0].transaction_id", is(successfulRefund.getExternalId()))
+                .body("results[1].transaction_id", is(submittedPayment.getExternalId()))
+                .body("results[2].transaction_id", is(nullValue()));
+    }
+
```
jankowiakmaria: In general regarding the tests it would read better if we asserted on count
``` .body("count", is(2))```
and then ids of results (but only the matching ones) rather than asserting on a null value for the third (not existing one)
```.body("results[2].transaction_id", is(nullValue()));```

In short replace ```.body("results[2].transaction_id", is(nullValue()));``` with ``` .body("count", is(2))```.

File: src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java [hunk 3]
```diff
@@ -304,6 +308,33 @@ private Long getOffset() {
         return offset;
     }
 
+    private List<String> getTransactionStatesForStatuses(CommaDelimitedSetParameter paymentStates) {
+        return paymentStates.getParameters().stream()
+                .map(TransactionState::getStatesForStatus)
+                .flatMap(List::stream)
+                .map(TransactionState::name)
+                .collect(Collectors.toList());
+    }
```
jankowiakmaria: Given this fancy stream logic, it would be good to assert on generated links as well. (or some sort of unit test for it)

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
mrlumbu: oh nice. I was being lazy

jankowiakmaria: it's just an improvement. The biggest thing was broken build.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
mrlumbu: the generated links need to just echo the input values, whereas this is mapping from input value to the TransactionState. So I'm not sure what could go wrong with links? Unless I'm misunderstanding you

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
mrlumbu: yep

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -274,7 +278,7 @@ public String buildQueryParamString(Long forPage) {
         if (isNotBlank(lastDigitsCardNumber)) {
             sb.append("&" + LAST_DIGITS_CARD_NUMBER_FIELD + "=" + lastDigitsCardNumber);
         }
-        if (paymentStates != null && !paymentStates.isEmpty()) {
+        if (isSet(paymentStates)) {
             sb.append("&" + PAYMENT_STATES_FIELD + "=" + paymentStates.getRawString());
         }
         if (refundStates != null && !refundStates.isEmpty()) {
```
jankowiakmaria: could use `isSet` method.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
jankowiakmaria: No, I misread. I thought it was being used in generating links as well :doh:

PR #244, opened 2019-08-20
BAU Make pact data static

mrlumbu:
Pact detects contract changes based on a hash of the contract content.
Therefore in order for semantically identical contracts not to be considered
different we need to write static data to the contract, rather than random
uuids etc.

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/util/fixture/QueueRefundEventFixture.java
+++ b/src/test/java/uk/gov/pay/ledger/util/fixture/QueueRefundEventFixture.java
@@ -4,11 +4,9 @@ import au.com.dius.pact.consumer.dsl.PactDslJsonBody;
 import com.amazonaws.services.sqs.AmazonSQS;
 import com.google.common.collect.ImmutableMap;
 import com.google.gson.GsonBuilder;
-import org.apache.commons.lang3.RandomStringUtils;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/ledger/util/fixture/QueueRefundEventFixture.java
```diff
@@ -4,24 +4,22 @@
...
 
 public class QueueRefundEventFixture implements QueueFixture<QueueRefundEventFixture, Event> {
     private String sqsMessageId;
     private ResourceType resourceType = ResourceType.REFUND;
     private Long amount = 50L;
     private String gatewayAccountId = "123456";
-    private String resourceExternalId = RandomStringUtils.randomAlphanumeric(20);
-    private String parentResourceExternalId = RandomStringUtils.randomAlphanumeric(20);
+    private String resourceExternalId = "resource_external_id";
+    private String parentResourceExternalId = "parent_resource_external_id";
```
jankowiakmaria: it's too long. It can be only 26 characters long (and is 27) therefore the build is failing.

mrlumbu: 🥇

PR #246, opened 2019-08-21
PP-5537 add hard coded values to some pact tests

SandorArpa:
## WHAT
- pact-js is not yet generating providerStates and thus we cannot pass params to be used on the provider side.


jankowiakmaria:
Does it mean it will be cleaned up once the params are being passed?

SandorArpa:
> Does it mean it will be cleaned up once the params are being passed?

It is, when pact-js is going to handle providerStates. There is a change request for it. Let's hope they are going to implement it soon.

PR #293, opened 2019-09-11
PP-5537 merge api pact tests

SandorArpa:
- merge the api pact tests into one file. This is needed to rid of PactFilter. Using PactFilter can stop (rightly) publications to the pact broker and failing to publish can break the pipeline.


src/test/java/uk/gov/pay/ledger/pact/TransactionsAndEventApiContractTest.java
```diff
@@ -145,6 +139,120 @@ public void createThreeTransactionsForSelfserviceSearch() {
...
+
+        aTransactionFixture()
+                .withExternalId(transactionExternalId)
+                .withGatewayAccountId(gatewayAccountId)
+                .withState(TransactionState.CREATED)
+                .withAmount(2000L)
+                .withCorporateCardSurcharge(250L)
+                .withTotalAmount(2250L)
+                .withCaptureSubmittedDate(ZonedDateTime.now())
+                .withCapturedDate(ZonedDateTime.now())
```
jankowiakmaria: It would be good to change to static dateTime instead of dynamic one.

PR #305, opened 2019-09-19
PP-5647: add exact_reference_match query param to search endpoint

jankowiakmaria:
Context:
Currently ledger searches based on partial reference (the `like` query).
The current behaviour of Connector though is to search by exact mathc for
public api calls and partial reference for selfservice.
In order to mimic the same behaviour in ledger the exact_reference_match query
needs to be introduced (with default false value) which will allow switching between
exact and partial match for reference.

Changes:
* add exact_reference_match to TransactionSearchParams (default true, not added to
query param string
* update getFilterTemplates and getQueryMap methods to support switching between exact
and partial mathch
* cover the changes with unit and integration tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
@@ -157,17 +157,17 @@ public class TransactionSearchParams {
             filters.add(" t.type = :" + TRANSACTION_TYPE_FIELD + "::transaction_type");
         }
         if (isNotBlank(email)) {
-            filters.add(" lower(t.email) ILIKE :" + EMAIL_FIELD);
+            filters.add(" lower(t.email) LIKE lower(:" + EMAIL_FIELD + ")");
         }
         if (isNotBlank(reference)) {
             if(exactReferenceMatch) {
                 filters.add(" lower(t.reference) = lower(:" + REFERENCE_FIELD + ")");
             } else {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -157,7 +160,11 @@ public void setDisplaySize(Long displaySize) {
             filters.add(" lower(t.email) ILIKE :" + EMAIL_FIELD);
         }
         if (isNotBlank(reference)) {
-            filters.add(" lower(t.reference) ILIKE :" + REFERENCE_FIELD);
+            if(exactReferenceMatch) {
+                filters.add(" lower(t.reference) = lower(:" + REFERENCE_FIELD + ")");
+            } else {
+                filters.add(" lower(t.reference) ILIKE :" + REFERENCE_FIELD);
```
kbottla: We could do `LIKE lower(reference)` which seems faster

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoSearchIT.java
```diff
@@ -137,6 +137,7 @@ public void shouldReturn1Record_whenSearchingByReference() {
 
         TransactionSearchParams searchParams = new TransactionSearchParams();
         searchParams.setAccountId(gatewayAccountId);
+        searchParams.setExactReferenceMatch(true);
```
kbottla: May be rename method name to `shouldReturn1Record_whenSearchingByExactReference`

PR #315, opened 2019-09-24
PP-5658 Report on card brand label to match Connector

sfount:
Connector responds with with the card brand human readable labels found
in Connector's database. This is now passed through the payment details
entered event and stored in the `card_brand_label` attribute of
transaction details. Send this in favour of the card brand key (which
can still be searched/ filtered as normal).


src/test/java/uk/gov/pay/ledger/transaction/model/TransactionFactoryTest.java
```diff
@@ -12,9 +12,7 @@
 import java.util.Optional;
 
 import static org.hamcrest.MatcherAssert.assertThat;
-import static org.hamcrest.Matchers.is;
-import static org.hamcrest.Matchers.notNullValue;
-import static org.hamcrest.Matchers.nullValue;
+import static org.hamcrest.Matchers.*;
```
kbottla: Wildcard import ?

PR #322, opened 2019-09-27
PP-5659 Add transaction live column

sfount:
* add boolean `live` column to transaction table
* index column for analytics/ searching

### DIFF
--- a/src/main/resources/migrations/00030_add_live_column_to_transaction_table.sql
+++ b/src/main/resources/migrations/00030_add_live_column_to_transaction_table.sql
@@ -5,5 +5,6 @@
 ALTER TABLE transaction
     ADD COLUMN live BOOLEAN;
 
---changeset uk.gov.pay:index_transaction_live_column
-CREATE INDEX transaction_live_idx ON transaction(live)
+--changeset uk.gov.pay:index_transaction_live_column runInTransaction:false
+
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00030_add_live_column_to_transaction_table.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_live_column_to_transaction_table
+
+ALTER TABLE transaction
+    ADD COLUMN live BOOLEAN;
+
+--changeset uk.gov.pay:index_transaction_live_column
+CREATE INDEX transaction_live_idx ON transaction(live)
```
kbottla:
```suggestion
CREATE INDEX CONCURRENTLY transaction_live_idx ON transaction(live)
```

kbottla: Not sure if postgres will use index effectively on low cardinality fields. Or we can create partial index

`CREATE INDEX CONCURRENTLY transaction_live_idx ON transaction(live) WHERE live is true;`

sfount: Great suggestion! A partial index would be much more useful to Postgres, have updated to do this and not run in transaction.

PR #333, opened 2019-10-02
PP-5671 Add report DAO

stephencdaly:
Add a reports DAO with a method to perform a query to filter payments and return the counts of the filtered payments in each state.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
+++ b/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
@@ -1,0 +1,60 @@
+package uk.gov.pay.ledger.report.dao;
+
+import org.apache.commons.lang3.StringUtils;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.statement.Query;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
```diff
@@ -0,0 +1,59 @@
...
+                    COUNT_TRANSACTIONS_BY_STATE,
+                    true);
+
+            Query query = handle.createQuery(template)
+                    .bind("transactionType", TransactionType.PAYMENT);
+            params.getQueryMap().forEach(query::bind);
+
+            return query.map((rs, rowNum) -> {
+                String state = rs.getString("state");
+                Long count = rs.getLong(2);
```
sfount: Could this be aliased as `count` in the query so that if further things are added to this it won't require the index to be correct?

stephencdaly: yes, good idea

stephencdaly:
As discussed, we'll restrict the query to live accounts when the live/test status is stored for payments and the database is fully migrated

PR #350, opened 2019-10-08
PP-5681 insert gateway_transaction_id when creating a payment type

SandorArpa:
## WHAT
- insert gateway_transaction_id when creating a telephone payment
- gateway transaction id is an important part of the telephone and thus we need to store it

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
@@ -35,6 +35,7 @@ public class TransactionDaoIT {
                 .withExternalMetadata(ImmutableMap.of("key1", "value1", "anotherKey", ImmutableMap.of("nestedKey", "value")))
                 .withTransactionType("PAYMENT")
                 .withLive(true)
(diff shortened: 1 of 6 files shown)


src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -35,6 +35,7 @@ public void shouldInsertTransaction() {
                 .withExternalMetadata(ImmutableMap.of("key1", "value1", "anotherKey", ImmutableMap.of("nestedKey", "value")))
                 .withTransactionType("PAYMENT")
                 .withLive(true)
+                .withGatewayTransactionId("gateway_transaction_id")
```
jankowiakmaria: [minor] Shouldn't we have an assertion that verifies this field too? (I can't see it there at the moment)

SandorArpa: It is going to be included. PR wasn't ready for review.

SandorArpa: No worries, as I said it's all my fault

PR #354, opened 2019-10-10
PP-5667 Return Card Type From Ledger

RoryMalcolm:
- Return the card type in lower case from ledger


src/main/java/uk/gov/pay/ledger/transaction/model/CardDetails.java
```diff
@@ -80,7 +92,8 @@ public boolean equals(Object o) {
                 Objects.equals(billingAddress, that.billingAddress) &&
                 Objects.equals(cardBrand, that.cardBrand) &&
                 Objects.equals(lastDigitsCardNumber, that.lastDigitsCardNumber) &&
-                Objects.equals(firstDigitsCardNumber, that.firstDigitsCardNumber);
+                Objects.equals(firstDigitsCardNumber, that.firstDigitsCardNumber) &&
+                Objects.equals(cardType, that.cardType);
```
stephencdaly: need to update `hashcode()`

RoryMalcolm: Done!

PR #359, opened 2019-10-15
PP-5602 Return 404 for invalid page number

kbottla:
## WHAT
- Currently we return 200 response and empty results, if `page number` query parameter is set to too high. To maintain functionality on par with connector/publicapi, return 404 when page number is greater than the last page number possible for query params

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
@@ -100,6 +100,15 @@ public class TransactionService {
 
         Long total = transactionDao.getTotalForSearch(searchParams);
 
+        long size = searchParams.getDisplaySize();
+        if (total > 0 && searchParams.getDisplaySize() > 0) {
+            long lastPage = (total + size - 1) / size;
+            if (searchParams.getPageNumber() > lastPage || searchParams.getPageNumber() < 1) {
+                throw new WebApplicationException("the requested page not found",
+                        Response.Status.NOT_FOUND);
+            }
+        }
+
         return buildTransactionSearchResponse(searchParams, uriInfo, transactionList, total);
     }
 
--- a/src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
@@ -160,6 +160,22 @@ public class TransactionServiceTest {
     }
 
     @Test
+    public void searchTransactionsWithoutParent_shouldThrowNotFoundException_forInvalidPaginationParams() {
+        List<TransactionEntity> transactionViewList = TransactionFixture.aTransactionList(gatewayAccountId, 10);
+        when(mockTransactionDao.searchTransactions(any(TransactionSearchParams.class))).thenReturn(transactionViewList);
+        when(mockTransactionDao.getTotalForSearch(any(TransactionSearchParams.class))).thenReturn(10L);
+        thrown.expect(WebApplicationException.class);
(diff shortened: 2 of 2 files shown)


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -100,6 +100,15 @@ private TransactionSearchResponse searchTransactionsWithoutParent(TransactionSea
 
         Long total = transactionDao.getTotalForSearch(searchParams);
 
+        long size = searchParams.getDisplaySize();
+        if (total > 0 && searchParams.getDisplaySize() > 0) {
+            long lastPage = (total + size - 1) / size;
```
jankowiakmaria: This formula is less intuitive [for me].
Why not `ceil(total/size)`?

I've found different approaches here: https://stackoverflow.com/questions/7139382/java-rounding-up-to-an-int-using-math-ceil.

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -100,6 +100,15 @@ private TransactionSearchResponse searchTransactionsWithoutParent(TransactionSea
...
+            if (searchParams.getPageNumber() > lastPage || searchParams.getPageNumber() < 1) {
```
jankowiakmaria: Shouldn't we have validation of `searchParams.getPageNumber() < 1` before we run any db queries?
(I'm trying to prematurely optimise ;) )

src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
```diff
@@ -159,6 +159,22 @@ public void shouldListTransactionsWithCorrectQueryParamsForPaginationLinks() {
...
 
+    @Test
+    public void searchTransactionsWithoutParent_shouldThrowNotFoundException_forInvalidPaginationParams() {
+        List<TransactionEntity> transactionViewList = TransactionFixture.aTransactionList(gatewayAccountId, 10);
+        when(mockTransactionDao.searchTransactions(any(TransactionSearchParams.class))).thenReturn(transactionViewList);
+        when(mockTransactionDao.getTotalForSearch(any(TransactionSearchParams.class))).thenReturn(10L);
+        thrown.expect(WebApplicationException.class);
+        thrown.expectMessage("the requested page not found");
+
+        searchParams.setPageNumber(100L);
```
jankowiakmaria: Wouldn't `2` be enough to cause the failure? (we have 10 transactions and  MAX_DISPLAY_SIZE of 500);
Ideally we should set the displaySize to 10 (or something) to test the edge case.

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
kbottla: We already have this in connector, didn't want to introduce something different
https://github.com/alphagov/pay-connector/blob/a6dde93f3c38f3272947fb606b053013dc223ae1/src/main/java/uk/gov/pay/connector/common/service/search/AbstractSearchStrategy.java#L21

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
kbottla: 👍 Will look into it separately

src/test/java/uk/gov/pay/ledger/transaction/service/TransactionServiceTest.java
kbottla: 👍 Updated

PR #366, opened 2019-10-18
PP-5170 Store New Event Details Data

RoryMalcolm:
- When we receive a new event from the queue, sometimes that event already
exists but with less information that the newer version of the event,
currently the behaviour of ledger is to not store this information, this
commit fixes that by updating the previous record with the new event
information if there is a disparity between the two.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/service/EventService.java
+++ b/src/main/java/uk/gov/pay/ledger/event/service/EventService.java
@@ -22,15 +22,6 @@ public class EventService {
         return EventDigest.fromEventList(events);
     }
 
-    public CreateEventResponse createIfDoesNotExist(Event event) {
-        try {
-            Optional<Long> status = eventDao.insertEventIfDoesNotExistWithResourceTypeId(event);
-            return new CreateEventResponse(status);
-        } catch (Exception e) {
-            return new CreateEventResponse(e);
-        }
-    }
-
     private CreateEventResponse updateExistingEvent(Event event) {
         try {
             return new CreateEventResponse(eventDao.updateIfExistsWithResourceTypeId(event));
--- a/src/main/java/uk/gov/pay/ledger/event/service/EventService.java
+++ b/src/main/java/uk/gov/pay/ledger/event/service/EventService.java
@@ -30,4 +30,21 @@ public class EventService {
             return new CreateEventResponse(e);
         }
     }
+
+    private CreateEventResponse updateExistingEvent(Event event) {
+        try {
+            return new CreateEventResponse(eventDao.updateIfExistsWithResourceTypeId(event));
+        } catch (Exception e) {
+            return new CreateEventResponse(e);
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/ledger/event/service/EventService.java
```diff
@@ -30,4 +30,21 @@ public CreateEventResponse createIfDoesNotExist(Event event) {
...
+
+    private CreateEventResponse updateExistingEvent(Event event) {
+        try {
+            return new CreateEventResponse(eventDao.updateIfExistsWithResourceTypeId(event));
+        } catch (Exception e) {
+            return new CreateEventResponse(e);
+        }
+    }
+
+    public CreateEventResponse createOrUpdateIfExists(Event event) {
```
stephencdaly: Looks like this replaces `createIfDoesNotExist`, so remove that

stephencdaly: Please add/alter tests in EventServiceTest

src/test/java/uk/gov/pay/ledger/event/service/EventServiceTest.java
```diff
@@ -110,21 +110,39 @@ public void createIfDoesNotExistReturnsSuccessfulCreatedResponse() {
...
-        CreateEventResponse response = eventService.createIfDoesNotExist(event);
+        CreateEventResponse response = eventService.createOrUpdateIfExists(event);
 
         assertFalse(response.isSuccessful());
         assertThat(response.getState(), is(CreateEventResponse.CreateEventState.ERROR));
         assertThat(response.getErrorMessage(), is("forced failure"));
     }
+
+    @Test
+    public void createORUpdateIfExistsUpdatesExistingEvent() {
```
stephencdaly: `createOR` make the R lowercase

src/test/java/uk/gov/pay/ledger/event/service/EventServiceTest.java
```diff
@@ -110,21 +110,39 @@ public void createIfDoesNotExistReturnsSuccessfulCreatedResponse() {
...
+                .toEntity();
+        Event newEvent = EventFixture.anEventFixture()
+                .withResourceExternalId(existingEvent.getResourceExternalId())
+                .withEventDate(existingEvent.getEventDate())
+                .withEventData("{\"key\": \"value\"}")
+                .toEntity();
+        when(mockEventDao.insertEventWithResourceTypeId(newEvent)).thenReturn(0L);
+        when(mockEventDao.updateIfExistsWithResourceTypeId(newEvent)).thenReturn(Optional.of(1L));
+        eventService.createOrUpdateIfExists(existingEvent);
+        CreateEventResponse response = eventService.createOrUpdateIfExists(newEvent);
```
stephencdaly: Don't think you need the 2 events and to call the method twice. As you've mocked out the dao it doesn't actually do anything. You can just use the `event` field rather than making events for this test and only call `createOrUpdateIfExists` once.

PR #380, opened 2019-10-24
PP-5427 add transactions summary

SandorArpa:
## WHAT

- add a method to get transactions summary for selfservice
- refactor ReportDao
- add relevant/refactor tests


src/main/java/uk/gov/pay/ledger/report/params/PaymentsReportParams.java
```diff
@@ -12,9 +12,7 @@
 
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
-public class PaymentsReportParams {
-    private static final String GATEWAY_ACCOUNT_EXTERNAL_FIELD = "account_id";
-    private static final String FROM_DATE_FIELD = "from_date";
+public class PaymentsReportParams implements ReportParams {
```
kbottla: We now have `PaymentsReportParams ` and `TransactionSummaryParams`. It may be confusing to know the usage / purpose for these with only a couple of differences. Fine for now, but we may need to revisit to consolidate into one.

PR #387, opened 2019-10-29
PP-5427 add net income to transaction summary

SandorArpa:
## WHAT
- add new field (net_income) to transaction summary
- update tests


src/main/java/uk/gov/pay/ledger/report/entity/TransactionSummaryResult.java
```diff
@@ -10,10 +10,14 @@
...
     private TransactionsStatisticsResult refunds;
+    private Long netIncome;
 
-    public TransactionSummaryResult(TransactionsStatisticsResult payments, TransactionsStatisticsResult refunds) {
+    public TransactionSummaryResult(TransactionsStatisticsResult payments,
+                                    TransactionsStatisticsResult refunds,
+                                    Long netIncome) {
         this.payments = payments;
         this.refunds = refunds;
+        this.netIncome = netIncome;
```
mrlumbu: surely this netIncome is something that the object can calculate? It doesn't have to be passed in?

PR #390, opened 2019-10-31
PP-5427 sum amount when total_amount is null

SandorArpa:
## WHAT
- sum amount when total_amount is null to get a precise statistical figure

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
+++ b/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
@@ -18,7 +18,9 @@ public class ReportDao {
             ":searchExtraFields " +
             "GROUP BY state";
 
-    private static final String TRANSACTION_SUMMARY_STATISTICS = "SELECT count(1) AS count, sum(amount) AS grossAmount FROM transaction t " +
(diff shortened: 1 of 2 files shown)


SandorArpa:
> Looks good to me 👍 neat solution!
> 
> It might be useful to run `EXPLAIN ANALYSE` on this query just to prove that for large data sets this will continue to be performant. That analysis shouldn't be needed for the self service use case but larger platform queries would need this.

select count(1), sum(case when total_amount is null then amount else total_amount end) from transaction; =>

"Aggregate  (cost=94.53..94.54 rows=1 width=40) (actual time=10.573..10.590 rows=1 loops=1)"
"  ->  Seq Scan on transaction  (cost=0.00..89.35 rows=1035 width=16) (actual time=0.114..5.460 rows=550 loops=1)"
"Planning time: 0.121 ms"
"Execution time: 10.690 ms"

select count(1), sum(amount) from transaction; =>

"Aggregate  (cost=94.53..94.54 rows=1 width=40) (actual time=11.099..11.115 rows=1 loops=1)"
"  ->  Seq Scan on transaction  (cost=0.00..89.35 rows=1035 width=8) (actual time=0.046..5.638 rows=550 loops=1)"
"Planning time: 0.144 ms"
"Execution time: 11.214 ms"

PR #396, opened 2019-11-05
PP-5779 Only project transaction if event inserted

sfount:
Move projecting transaction condition logic to the `EventMessageHandler` service and out of the DAO.

Advantages:
* don't rely on de-normalised `event_count`
* move business logic out of the DAO, these should be easier to
understand
* group all writing transaction logic into one place


sfount:
We've decided keeping this the way it is is safest, it has worked so far. This will require updates to the event count column in the production database (given a prior issue with updating events) but that's less than thinking through all new behaviours with this.

PR #408, opened 2019-11-12
PP-5770 add new state to contract tests to include all fields

SandorArpa:
## WHAt
- add a new state that has a payment with all fields and a corresponding refund with parent transaction

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/pact/ContractTest.java
+++ b/src/test/java/uk/gov/pay/ledger/pact/ContractTest.java
@@ -320,6 +320,23 @@ public abstract class ContractTest {
                 .insert(app.getJdbi());
     }
 
+    @State("a payment with all fields and a corresponding refund exists")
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/ledger/pact/ContractTest.java
```diff
@@ -375,4 +394,30 @@ private void createPaymentTransactionForSelfserviceSearch(String transactionExte
...
+        aTransactionFixture()
+                .withExternalId(transactionExternalId)
+                .withGatewayAccountId(gatewayAccountId)
+                .withTransactionType(TransactionType.PAYMENT.name())
+                .withAmount(1000L)
+                .withReference(reference)
+                .withState(state)
+                .withDefaultCardDetails()
+                .withFirstDigitsCardNumber("424242")
+                .withLastDigitsCardNumber("4242")
```
alexbishop1: Would it be possible to make the last digits not be a subset of the first digits? It could get confusing or allow bugs to sneak past if we accidentally grab from the wrong end of the card number.

PR #445, opened 2019-12-10
PP-5893-Add-MDC-Logging-Filter

nimalank7:
Description:
- At the moment this doesn't compile as Pay Java Commons doesn't have the static members PARENT_EXTERNAL_ID and TRANSACTION_EXTERNAL_ID
- Adds eventId, parent_external_id and transactionExternalId to MDC

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
+++ b/src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
@@ -1,0 +1,32 @@
+package uk.gov.pay.ledger.filters;
+
+import org.slf4j.MDC;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerRequestFilter;
+import java.io.IOException;
+import java.util.Optional;
+
+import static uk.gov.pay.logging.LoggingKeys.LEDGER_EVENT_ID;
+
+public class LoggingMDCRequestFilter implements ContainerRequestFilter {
+
+    public static final String PARENT_TRANSACTION_EXTERNAL_ID = "parent_transaction_external_id";
+    public static final String TRANSACTION_EXTERNAL_ID = "transaction_external_id";
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        getPathParameterFromRequest("eventId", requestContext)
+                .ifPresent(eventId -> MDC.put(LEDGER_EVENT_ID, eventId));
+
+        getPathParameterFromRequest("transactionExternalId", requestContext)
+                .ifPresent(transactionExternalId -> MDC.put(TRANSACTION_EXTERNAL_ID, transactionExternalId));
+
+        getPathParameterFromRequest("parentTransactionExternalId", requestContext)
+                .ifPresent(parentTransactionExternalId -> MDC.put(PARENT_TRANSACTION_EXTERNAL_ID, parentTransactionExternalId));
+    }
+
+    private Optional<String> getPathParameterFromRequest(String parameterName, ContainerRequestContext requestContext) {
+        return Optional.ofNullable(requestContext.getUriInfo().getPathParameters().getFirst(parameterName));
+    }
+}
--- a/src/main/java/uk/gov/pay/ledger/filters/LoggingMDCResponseFilter.java
+++ b/src/main/java/uk/gov/pay/ledger/filters/LoggingMDCResponseFilter.java
@@ -1,0 +1,16 @@
+package uk.gov.pay.ledger.filters;
+
+import org.slf4j.MDC;
+
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import java.io.IOException;
+
+public class LoggingMDCResponseFilter implements ContainerResponseFilter {
+
+    @Override
+    public void filter(ContainerRequestContext containerRequestContext, ContainerResponseContext containerResponseContext) throws IOException {
+        MDC.clear();
+    }
+}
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
```diff
@@ -0,0 +1,36 @@
...
+import java.io.IOException;
+import java.util.Optional;
+
+import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
+import static uk.gov.pay.logging.LoggingKeys.LEDGER_EVENT_ID;
+
+public class LoggingMDCRequestFilter implements ContainerRequestFilter {
+
+    public static final String PARENT_EXTERNAL_ID = "parent_external_id";
+    public static final String TRANSACTION_EXTERNAL_ID = "transactionExternalId";
```
stephencdaly: all the keys we log should be snake_case

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.ws.rs.container.ContainerRequestFilter;
+import java.io.IOException;
+import java.util.Optional;
+
+import static uk.gov.pay.logging.LoggingKeys.GATEWAY_ACCOUNT_ID;
+import static uk.gov.pay.logging.LoggingKeys.LEDGER_EVENT_ID;
+
+public class LoggingMDCRequestFilter implements ContainerRequestFilter {
+
+    public static final String PARENT_EXTERNAL_ID = "parent_external_id";
```
stephencdaly: might be good to make this key "parent_transaction_external_id" so it's more obvious what it refers too

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCResponseFilter.java
```diff
@@ -0,0 +1,16 @@
...
+import javax.ws.rs.container.ContainerRequestContext;
+import javax.ws.rs.container.ContainerResponseContext;
+import javax.ws.rs.container.ContainerResponseFilter;
+import java.io.IOException;
+
+public class LoggingMDCResponseFilter implements ContainerResponseFilter {
+
+    @Override
+    public void filter(ContainerRequestContext containerRequestContext, ContainerResponseContext containerResponseContext) throws IOException {
+        MDC.clear();
```
stephencdaly: What's the reason why in some of the apps we specify which MDC keys to remove? I'm not familiar with this stuff.

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
oswaldquek: this will need to be `transaction_external_id` to follow the GDS json logging convention of snake case for keys.

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
```diff
@@ -0,0 +1,36 @@
...
+    public static final String TRANSACTION_EXTERNAL_ID = "transactionExternalId";
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        getPathParameterFromRequest("accountId", requestContext)
+                .ifPresent(accountId -> MDC.put(GATEWAY_ACCOUNT_ID, accountId));
```
oswaldquek: i don't think we want to do this as "accountId" is a query parameter for some paths, but they're never a path parameter.

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
```diff
@@ -0,0 +1,36 @@
...
+
+        getPathParameterFromRequest("eventId", requestContext)
+                .ifPresent(eventId -> MDC.put(LEDGER_EVENT_ID, eventId));
+
+        getPathParameterFromRequest("parent_external_id", requestContext)
```
oswaldquek: this should be "parentExternalId" as per https://github.com/alphagov/pay-ledger/blob/master/src/main/java/uk/gov/pay/ledger/transaction/resource/TransactionResource.java#L106

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCResponseFilter.java
nimalank7: I'm not sure either - in publicauth the only key logged there is account_id so we can just specify to remove that. I'm pretty sure we have to clear all of the MDC keys as its thread local.

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
nimalank7: Done.

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCResponseFilter.java
oswaldquek: yeah we could have probably used MDC.clear everywhere.

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
stephencdaly: Sorry I meant the other one, we should have `TRANSACTION_EXTERNAL_ID` and `PARENT_TRANSACTION_EXTERNAL_ID` I think

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
nimalank7: Done

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
nimalank7: Done

src/main/java/uk/gov/pay/ledger/filters/LoggingMDCRequestFilter.java
nimalank7: Done

PR #478, opened 2019-12-31
PP-5978: Performance DAO for all payment transactions

oswaldquek:
This commit is just for the Performance DAO, not the resource level class(es).

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
+++ b/src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
@@ -1,0 +1,28 @@
+package uk.gov.pay.ledger.performance.dao;
+
+import org.jdbi.v3.core.Jdbi;
+import uk.gov.pay.ledger.performance.dao.mapper.PerformanceReportEntityMapper;
+import uk.gov.pay.ledger.performance.entity.PerformanceReportEntity;
+
+import javax.inject.Inject;
+
+public class PerformanceDao {
+
+    private static final String PERFORMANCE_REPORT =
+            "select coalesce(count(amount),0) as volume, coalesce(sum(amount),0) total_amount, " +
+                    "coalesce(avg(amount),0) avg_amount from transaction " +
+            "where state='SUCCESS' and type='PAYMENT' AND  transaction_details @> '{\"live\": true}'";
+
+    private final Jdbi jdbi;
+
+    @Inject
+    public PerformanceDao(Jdbi jdbi) {
+        this.jdbi = jdbi;
+    }
+
+    public PerformanceReportEntity performanceReportForPaymentTransactions() {
+        return jdbi.withHandle(handle ->
+                handle.createQuery(PERFORMANCE_REPORT)
+                        .map(new PerformanceReportEntityMapper()).findOnly());
+    }
+}
--- a/src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
+++ b/src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
@@ -11,7 +11,7 @@ public class PerformanceDao {
     private static final String PERFORMANCE_REPORT =
             "select coalesce(count(amount),0) as volume, coalesce(sum(amount),0) total_amount, " +
                     "coalesce(avg(amount),0) avg_amount from transaction " +
-            "where state='SUCCESS' and type='PAYMENT' AND  transaction_details @> '{\"live\": true}'";
+            "where state='SUCCESS' and type='PAYMENT' AND live=true";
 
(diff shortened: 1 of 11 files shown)


src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
```diff
@@ -0,0 +1,28 @@
...
+import uk.gov.pay.ledger.performance.entity.PerformanceReportEntity;
+
+import javax.inject.Inject;
+
+public class PerformanceDao {
+
+    private static final String PERFORMANCE_REPORT =
+            "select coalesce(count(amount),0) as volume, coalesce(sum(amount),0) total_amount, " +
+                    "coalesce(avg(amount),0) avg_amount from transaction " +
+            "where state='SUCCESS' and type='PAYMENT' AND  transaction_details @> '{\"live\": true}'";
```
sfount: This should use the `live` column on the transaction directly -- having a condition on a non-indexed attribute would end up costing a lot.

The `live` column is populated from this value when the transaction is made (and has a partial index https://github.com/alphagov/pay-ledger/pull/322)

src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
```diff
@@ -0,0 +1,28 @@
+package uk.gov.pay.ledger.performance.dao;
```
sfount: I think this could all live under the 'report' package, all of these routes will provide higher level aggregate info and having it all in the same place seems to make sense.

Either this could be part of the `ReportDao` directly or it could be a new Dao still under the report package.

src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
oswaldquek: Oh! I missed the "live" column.

jankowiakmaria: There is a `live` indexed column in `transaction`.

src/main/java/uk/gov/pay/ledger/performance/dao/PerformanceDao.java
```diff
@@ -0,0 +1,28 @@
...
+import uk.gov.pay.ledger.performance.dao.mapper.PerformanceReportEntityMapper;
+import uk.gov.pay.ledger.performance.entity.PerformanceReportEntity;
+
+import javax.inject.Inject;
+
+public class PerformanceDao {
+
+    private static final String PERFORMANCE_REPORT =
+            "select coalesce(count(amount),0) as volume, coalesce(sum(amount),0) total_amount, " +
+                    "coalesce(avg(amount),0) avg_amount from transaction " +
```
jankowiakmaria: [style]
I'd like to see consistency in the query - either use `AS alias` for all of the columns or for none of them (instead of mixing both ways of aliasing columns) - no difference in terms of validity of the query

src/test/java/uk/gov/pay/ledger/performance/PerformanceDaoIT.java
```diff
@@ -0,0 +1,61 @@
...
+        PerformanceReportEntity performanceReport = transactionDao.performanceReportForPaymentTransactions();
+        assertThat(performanceReport.getTotalVolume(), is(3L));
+        assertThat(performanceReport.getTotalAmount(), Matchers.is(closeTo(expectedTotalAmount, ZERO)));
+        assertThat(performanceReport.getAverageAmount(), Matchers.is(closeTo(expectedAverageAmount, ZERO)));
+    }
+
+    private static String toJsonString(Map<String, ?> map) {
+        return gson.toJson(map);
+    }
+}
```
jankowiakmaria: [minor] Valid test - feels slightly overcomplicated to me.
I prefer explicit assertions (instead of calculating the `expectedAverageAmount` and  `expectedTotalAmount`).
I'd rather add few easy to calculate avg values and set the value explicitly in the assertion.

oswaldquek: Are there any reasons for the preference of explicit assertions?
I like the way its done here as I can see the expected values are calculated from the `relevantAmounts` list. This way I can easily deduce that the first three transactions that are inserted into the test are irrelevant.

jankowiakmaria: test should not (or rarely should) do any calculations - it increases the risk of testing the code with the same code (for example: testing the average function with the assertion that uses exactly the same function to calculate the expected value).
Explicit is easier to read.
Good names are 👍, but explicit assertions are easier to read and verify (imho).

PR #479, opened 2019-12-31
PP-5978: Implement /v1/report/performance-report endpoint

oswaldquek:
I saw no need to implement a Service class here, so the PerformanceReportResource calls the PerformanceReportDao directly.


src/main/java/uk/gov/pay/ledger/report/entity/PerformanceReportEntity.java
```diff
@@ -1,10 +1,18 @@
 package uk.gov.pay.ledger.report.entity;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import java.math.BigDecimal;
 
 public class PerformanceReportEntity {
+
+    @JsonProperty("total_volume")
```
jankowiakmaria: [Minor]
```java
@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
```
could be used at the class level instead of defining property names explicitly.

PR #480, opened 2019-12-31
PP-5991 Events ticker for dashboard

sfount:
"Administrative" list all events endpoint - used to drive a payment events ticker.

This will fetch events within a specified timeframe selecting only for `live` transactions. 

Future improvements: 
* Allow selecting by individual gateway accounts 
* Limit the consumer from selecting a large number of events (or a long timespan) 

These events can then be formatted and displayed as payments are processed (rough draft):

Screen Shot 2019-12-31 at 18 07 17

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/dao/EventDao.java
+++ b/src/main/java/uk/gov/pay/ledger/event/dao/EventDao.java
@@ -10,13 +10,17 @@ import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import org.jdbi.v3.sqlobject.statement.SqlUpdate;
 import org.jdbi.v3.sqlobject.transaction.Transaction;
 import uk.gov.pay.ledger.event.dao.mapper.EventMapper;
+import uk.gov.pay.ledger.event.dao.mapper.EventTickerMapper;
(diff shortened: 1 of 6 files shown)


sfount:
`EXPLAIN ANALYSE` for the proposed query: 
```
Nested Loop  (cost=0.86..39.73 rows=3 width=99) (actual time=0.022..0.029 rows=2 loops=1)
   ->  Index Scan Backward using event_date_idx on event  (cost=0.43..14.40 rows=3 width=56) (actual time=0.004..0.006 rows=2 loops=1)
         Index Cond: ((event_date >= '2020-01-02 14:50:00+00'::timestamp with time zone) AND (event_date <= '2020-01-02 14:55:00+00'::timestamp with time zone))
   ->  Index Scan using external_id_key on transaction  (cost=0.42..8.44 rows=1 width=911) (actual time=0.008..0.008 rows=1 loops=2)
         Index Cond: ((external_id)::text = (event.resource_external_id)::text)
         Filter: (NOT live)
 Planning Time: 0.324 ms
 Execution Time: 0.050 ms
```

It looks like only rows within the time period are inspected given the `btree` index, this should mean the query is always performant.

PR #481, opened 2020-01-02
PP-5970 Use Map for CSV resource endpoint

kbottla:
## WHAT
- Replaces POJO with Map<String,Object> for CSV data endpoint response
- Few fixes in CsvTransactionFactory  data mapping

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/util/csv/CSVMessageBodyWriter.java
+++ b/src/main/java/uk/gov/pay/ledger/util/csv/CSVMessageBodyWriter.java
@@ -1,5 +1,6 @@
 package uk.gov.pay.ledger.util.csv;
 
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.dataformat.csv.CsvMapper;
 import com.fasterxml.jackson.dataformat.csv.CsvSchema;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/ledger/util/csv/CSVMessageBodyWriter.java
```diff
@@ -36,11 +38,16 @@ public void writeTo(List data, Class aClass, Type type, Annotation[] annotations
...
             CsvMapper mapper = new CsvMapper();
 
-            // rank properties by class property order instead of alphabetically
             mapper.disable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY);
-            Object sample = data.get(0);
-            CsvSchema schema = mapper.schemaFor(sample.getClass()).withHeader();
+            mapper.configure(JsonGenerator.Feature.IGNORE_UNKNOWN, true);
+
+            CsvSchema.Builder builder = CsvSchema.builder();
+            Map<String, Object> headerRow = (Map<String, Object>) data.get(data.size() - 1); // get last record which is header
```
jankowiakmaria: Feels a bit hacky.
It would be cleaner to have an object having headers separate from data (two different properties) rather than having this assumption that the last record is a header (someone can change it by accident and the whole report looks interesting).

kbottla: Agree, but can't have headers separately as we can only pass `data` list to CSVMessageBodyWriter and as we need to generate headers from data, headers are tied to data set.

Other way is to explore other libraries for CSV data generation, but as discussed we would like to see the performance when using Maps. With the POJO we could get up 35K records successfully and failed to get 70K odd.

PR #482, opened 2020-01-02
PP-5978: Performance report by date range

oswaldquek:


src/main/java/uk/gov/pay/ledger/report/resource/PerformanceReportResource.java
```diff
@@ -25,7 +28,12 @@ public PerformanceReportResource(PerformanceReportDao performanceReportDao) {
     @Path("/performance-report")
     @GET
     @Timed
-    public PerformanceReportEntity getPerformanceReport() {
-        return performanceReportDao.performanceReportForPaymentTransactions();
+    public PerformanceReportEntity getPerformanceReport(@QueryParam("from_date") String fromDate,
+                                                        @QueryParam("to_date") String toDate) {
+
+        if (isBlank(fromDate) && isBlank(toDate)) {
```
kbottla: Should this be `||` ? if at least one of fromDate & toDate is not null, date range will be applied and `between` will have null comparison.

oswaldquek: Not quite. This PR is for "Scenario: retrieve the performance report for the payment transactions for given date range". 
There will be another PR for "Scenario: retrieve the performance report for the payment transactions specifying only from_date or to_date" which is what you're referring to :)

PR #483, opened 2020-01-02
PP-5978: Validation for date range

oswaldquek:


src/main/java/uk/gov/pay/ledger/report/resource/PerformanceReportResource.java
```diff
@@ -31,9 +33,13 @@ public PerformanceReportResource(PerformanceReportDao performanceReportDao) {
     public PerformanceReportEntity getPerformanceReport(@QueryParam("from_date") String fromDate,
                                                         @QueryParam("to_date") String toDate) {
 
-        if (isBlank(fromDate) && isBlank(toDate)) {
-            return performanceReportDao.performanceReportForPaymentTransactions();
+        if ((isNotBlank(fromDate) && isBlank(toDate)) || (isNotBlank(toDate) && isBlank(fromDate))) {
```
sfount: Is not the equivalent of 
```java
if (isBlank(fromDate) || isBlank(toDate)) { 
}
```

If I haven't misunderstood this might be simpler to read.

oswaldquek: `if (isBlank(fromDate) || isBlank(toDate))` would mean the request without any date parameters (which is a valid scenario) will not work. I made this mistake too - but the tests caught it!

sfount: Oh this version would allow you to not specify either -- in that case this doesn't do exactly the same thing. 

It would have to be two ifs to first check either then check both. That's not as clear an improvement.

sfount: Yeah the equivalent would be:
```java
boolean dateParamsProvided = isNotBlank(fromDate) || isNotBlankDate)
if (dateParamsProvided) { 
  if (isBlank(fromDate) || isBlank(toDate)) {
    // throw exception
  }
}
```

Certainly more verbose but potentially more readable too

PR #485, opened 2020-01-02
PP-5991 Timeseries transaction volumes by hour

sfount:
Transaction volumes grouped by hour will be used to show payment trends for a single day or comparing multiple days. Transaction volume data is grouped using the `created_date` field.

This end point requires from and to date to be specified - it should potentially limit the consumer to requesting within a one day period for performance reasons.

Future improvements: 
* optionally select data by gateway account or provide the override flag
* limit the consumer from doing huge massive queries

This will then be presented in a graph (rough draft)

Screen Shot 2020-01-02 at 12 04 13

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
+++ b/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
@@ -3,15 +3,21 @@ package uk.gov.pay.ledger.report.dao;
 import org.apache.commons.lang3.StringUtils;
 import org.jdbi.v3.core.Jdbi;
 import org.jdbi.v3.core.statement.Query;
+import org.jdbi.v3.sqlobject.config.RegisterRowMapper;
+import org.jdbi.v3.sqlobject.statement.SqlQuery;
 import uk.gov.pay.ledger.report.entity.PaymentCountByStateResult;
+import uk.gov.pay.ledger.report.entity.TimeseriesReportSlice;
 import uk.gov.pay.ledger.report.entity.TransactionsStatisticsResult;
+import uk.gov.pay.ledger.report.mapper.ReportMapper;
 import uk.gov.pay.ledger.report.params.TransactionSummaryParams;
 import uk.gov.pay.ledger.transaction.model.TransactionType;
 import uk.gov.pay.ledger.transaction.state.TransactionState;
 
 import javax.inject.Inject;
+import java.time.ZonedDateTime;
 import java.util.List;
 
+@RegisterRowMapper(ReportMapper.class)
 public class ReportDao {
     private static final String COUNT_TRANSACTIONS_BY_STATE = "SELECT state, count(1) AS count FROM transaction t " +
             "WHERE type = :transactionType::transaction_type " +
@@ -64,6 +70,26 @@ public class ReportDao {
                 return new TransactionsStatisticsResult(count, grossAmount);
             }).findOnly();
         });
+    }
+
+    // @TODO(sfount) using an interface with annotations seems like a neater way of writing a DAO (see events)
+    // @TODO(sfount) selecting this report by individual gateway account would be useful in the future
+    public List<TimeseriesReportSlice> getTransactionsVolumeByTimeseries(ZonedDateTime fromDate, ZonedDateTime toDate) {
+        return jdbi.withHandle(handle -> handle.createQuery("SELECT " +
+                    "date_trunc('hour', t.created_date) as timestamp, " +
+                    "COUNT(1) as all_payments, " +
+                    "COUNT(1) filter (WHERE t.state IN ('ERROR', 'GATEWAY_ERROR')) as errored_payments, " +
+                    "COUNT(1) filter (WHERE t.state IN ('SUCCESS')) as completed_payments, " +
+                    "SUM(t.amount) as amount, SUM(t.net_amount) as net_amount, SUM(t.total_amount) as total_amount, SUM(t.fee) as fee " +
+                    "FROM transaction t " +
+                    "WHERE t.live AND t.created_date >= :fromDate AND t.created_date <= :toDate " +
+                    "GROUP BY date_trunc('hour', t.created_date) " +
+                    "ORDER BY date_trunc('hour', t.created_date)")
+                    .bind("fromDate", fromDate)
+                    .bind("toDate", toDate)
+                    .map(new ReportMapper())
+                    .list()
+        );
     }
 
     private String createSearchTemplate(
--- a/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
+++ b/src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
@@ -74,12 +74,12 @@ public class ReportDao {
     public List<TimeseriesReportSlice> getTransactionsVolumeByTimeseries(ZonedDateTime fromDate, ZonedDateTime toDate) {
         return jdbi.withHandle(handle -> handle.createQuery("SELECT " +
                     "date_trunc('hour', t.created_date) as timestamp, " +
-                    "COUNT(1) as all_payments, " +
-                    "COUNT(1) filter (WHERE t.state IN ('ERROR', 'ERROR_GATEWAY')) as errored_payments, " +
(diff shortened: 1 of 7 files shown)


sfount:
Briefly looked at the query plan for `EXPLAIN ANALYSE` -- for two days in `test-12`
```
GroupAggregate  (cost=3493.84..3538.64 rows=896 width=160) (actual time=0.144..0.164 rows=5 loops=1)
   Group Key: (date_trunc('hour'::text, created_date))
   ->  Sort  (cost=3493.84..3496.08 rows=896 width=48) (actual time=0.132..0.136 rows=75 loops=1)
         Sort Key: (date_trunc('hour'::text, created_date))
         Sort Method: quicksort  Memory: 30kB
         ->  Bitmap Heap Scan on transaction  (cost=57.61..3449.90 rows=896 width=48) (actual time=0.026..0.114 rows=75 loops=1)
               Recheck Cond: ((created_date > '2020-01-01 00:00:00+00'::timestamp with time zone) AND (created_date < '2020-01-02 23:59:59+00'::timestamp with time zone))
               Filter: (type = 'PAYMENT'::transaction_type)
               Heap Blocks: exact=36
               ->  Bitmap Index Scan on created_date_idx  (cost=0.00..57.38 rows=896 width=0) (actual time=0.017..0.017 rows=273 loops=1)
                     Index Cond: ((created_date > '2020-01-01 00:00:00+00'::timestamp with time zone) AND (created_date < '2020-01-02 23:59:59+00'::timestamp with time zone))
 Planning Time: 0.203 ms
 Execution Time: 0.204 m
```

As this only looked at ~800 rows it looks like it is correctly making use of indexes on `created_date` -- this should mean it scales reasonably.

src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
```diff
@@ -66,6 +71,24 @@ public TransactionsStatisticsResult getTransactionSummaryStatistics(TransactionS
         });
     }
 
+    public List<TimeseriesReportSlice> getTransactionsVolumeByTimeseries(ZonedDateTime fromDate, ZonedDateTime toDate) {
+        return jdbi.withHandle(handle -> handle.createQuery("SELECT " +
+                    "date_trunc('hour', t.created_date) as timestamp, " +
+                    "COUNT(1) as all_payments, " +
```
oswaldquek: we `count(*)` not be used instead? it has better performance than `count(1)`

src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
```diff
@@ -66,6 +71,24 @@ public TransactionsStatisticsResult getTransactionSummaryStatistics(TransactionS
...
+                    "COUNT(1) filter (WHERE t.state IN ('ERROR', 'ERROR_GATEWAY')) as errored_payments, " +
+                    "COUNT(1) filter (WHERE t.state IN ('SUCCESS')) as completed_payments, " +
+                    "SUM(t.amount) as amount, SUM(t.net_amount) as net_amount, SUM(t.total_amount) as total_amount, SUM(t.fee) as fee " +
+                    "FROM transaction t " +
+                    "WHERE t.live AND t.created_date >= :fromDate AND t.created_date <= :toDate " +
```
oswaldquek: `t.created_date between :fromDate and :toDate`  is more concise

src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
sfount: Curious, this (quite in depth) analysis agrees with that but we've had issues with `COUNT(*)` before -- I wonder if it was just because indexes were being incorrectly used. Happy to change it!

src/main/java/uk/gov/pay/ledger/report/resource/ReportResource.java
```diff
@@ -69,4 +75,20 @@ public TransactionSummaryResult getTransactionSummaryResult(
...
     }
+
+    @Path("/transactions-by-hour")
+    @GET
+    @Timed
+    public List<TimeseriesReportSlice> getTransactionsByHour(
+            @QueryParam("from_date") String fromDate,
+            @QueryParam("to_date") String toDate
+    ) {
+        if(isBlank(fromDate) || isBlank(toDate)) {
```
oswaldquek: i know this works but it might be good to have a test for this validation

src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
sfount: Oh it was the ID itself the code was counting before -- this shows the fairly significant performance improvement https://github.com/alphagov/pay-ledger/pull/259

src/test/java/uk/gov/pay/ledger/report/dao/ReportDaoIT.java
```diff
@@ -378,4 +379,83 @@ public void shouldSumAmountWhenTotalAmountIsNull() {
...
+                .toEntity();
+
+        List<TimeseriesReportSlice> timeseriesReportSlices = reportDao.getTransactionsVolumeByTimeseries(
+                ZonedDateTime.parse("2019-09-30T00:00:00.000Z"),
+                ZonedDateTime.parse("2019-09-30T23:59:59.999Z")
+        );
+        assertThat(timeseriesReportSlices.size(), is(4));
+
+        assertThat(timeseriesReportSlices.get(0).getTimestamp().getHour(), is(0));
+        assertThat(timeseriesReportSlices.get(0).getAllPayments(), is(1));
```
oswaldquek: do we not care about amount, total amount, completed payments and errored payments here?

src/test/java/uk/gov/pay/ledger/report/dao/ReportDaoIT.java
```diff
@@ -378,4 +379,83 @@ public void shouldSumAmountWhenTotalAmountIsNull() {
...
+
+        assertThat(timeseriesReportSlices.get(1).getTimestamp().getHour(), is(8));
+        assertThat(timeseriesReportSlices.get(1).getAmount(), is(2000));
+        assertThat(timeseriesReportSlices.get(1).getTotalAmount(), is(3000));
+        assertThat(timeseriesReportSlices.get(1).getAllPayments(), is(2));
+        assertThat(timeseriesReportSlices.get(1).getCompletedPayments(), is(1));
+        assertThat(timeseriesReportSlices.get(1).getErroredPayments(), is(0));
+
+        assertThat(timeseriesReportSlices.get(2).getTimestamp().getHour(), is(9));
+        assertThat(timeseriesReportSlices.get(2).getAllPayments(), is(1));
```
oswaldquek: same comment as above

src/test/java/uk/gov/pay/ledger/report/dao/ReportDaoIT.java
sfount: I don't think so -- each group is checking a different scenario/ angle without asserting on every single value. So it's assuming that completed/ errored payments are covered somewhere else. Hopefully that's the case 👀

oswaldquek: Hmm maybe I don't really understand what you're trying to say, but from a java objects point of view won't be there be values for the fields in the `timeseriesReportSlices.get(0)` object? Here the completed payments field has the value of 1, do we not care about it?

src/main/java/uk/gov/pay/ledger/report/dao/ReportDao.java
oswaldquek: i'm not really an expert here, i just made the original comment based on some quick searching. i'm happy to settle for whatever you think is best here!

src/test/java/uk/gov/pay/ledger/report/dao/ReportDaoIT.java
sfount: Yeah poorly explained by me! Just trying to strike a balance between covering the functionality of the query and not being overly verbose or restrictive to avoid flake. I'll add some extra assertions to cover what you've mentioned

sfount:
@oswaldquek https://github.com/alphagov/pay-ledger/pull/485/commits/0160f1f88a7601493380e30493ed982cfc757faf

PR #493, opened 2020-01-03
PP-5978: Add state query param to endpoint

oswaldquek:


src/main/java/uk/gov/pay/ledger/report/resource/PerformanceReportResource.java
```diff
@@ -32,21 +34,36 @@ public PerformanceReportResource(PerformanceReportDao performanceReportDao) {
...
+                                                        @QueryParam("state") String state) {
 
+        var params = addDateRangeParamsOrThrow(PerformanceReportParamsBuilder.builder(), fromDate, toDate);
+        params = addStateParamOrThrow(params, state);
+        return performanceReportDao.performanceReportForPaymentTransactions(params.build());
+    }
+
+    private PerformanceReportParamsBuilder addStateParamOrThrow(PerformanceReportParamsBuilder builder, String state) {
+        if (isNotBlank(state)) {
+            Optional.ofNullable(TransactionState.from(state)).ifPresentOrElse(
```
sfount: Very minor but I think there is an `orElseThrow` on the `Optional` construct, that might ready nicely. 

Something like: 

```java 
Optional.ofNullable(TransactionState.from(state))
  .map(builder::withState)
  .orElseThrow(() -> new ValidationException("..."))
```

oswaldquek: This suggestion is better; I've updated the code

PR #498, opened 2020-01-03
PP-5970 Use BufferedWriter for CSV data

kbottla:
## WHAT
- Use BufferedWriter to write each line individually to output stream for CSV rather than streaming entire content.


kbottla:
Thanks @oswaldquek . Resource tests exists for both headers (https://github.com/alphagov/pay-ledger/blob/cd5ea2cea1f0ca2d8bf6778deb0f20613a5aceac/src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java#L695) and data (https://github.com/alphagov/pay-ledger/blob/cd5ea2cea1f0ca2d8bf6778deb0f20613a5aceac/src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java#L613)

PR #505, opened 2020-01-07
BAU - Downgrade Maven failsafe to 3.0.0-M1

kbottla:
## WHAT
- Downgrade Maven failsafe plugin as latest version doesn't run integration tests on `mvn clean verify`
- Fixed few failing integration tests
- Run integration tests on master

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -319,13 +319,8 @@
             </plugin>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <version>${surefire.version}</version>
-            </plugin>
-            <plugin>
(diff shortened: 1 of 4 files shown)


pom.xml
```diff
@@ -317,15 +317,10 @@
                     </execution>
                 </executions>
             </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-surefire-plugin</artifactId>
-                <version>${surefire.version}</version>
-            </plugin>
```
alexbishop1: Why was this removed?

kbottla: We already define surefire-plugin in profiles (default & contract-tests) and this is redundant

PR #523, opened 2020-01-15
PP-6013 Store Source in Transaction Column

rorymalcolm:
- When a TransactionEntity is serialised to the database, the source
value is stored in the source column within the transaction column

- When a TransactionEntity is deserialised from the the TransactionDao
reading the database, the source column's value is correctly read

- Tests are updated, TransactionDaoIT and TransactionEntityFactoryTest
to ensure this behaviour is as expected

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/event/model/TransactionEntityFactoryTest.java
+++ b/src/test/java/uk/gov/pay/ledger/event/model/TransactionEntityFactoryTest.java
@@ -7,6 +7,7 @@ import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
+import uk.gov.pay.commons.model.Source;
 import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
 
 import java.io.IOException;
@@ -14,6 +15,8 @@ import java.util.List;
 import java.util.Map;
 
 import static org.hamcrest.CoreMatchers.is;
+import static org.hamcrest.CoreMatchers.notNullValue;
+import static org.hamcrest.CoreMatchers.nullValue;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static uk.gov.pay.ledger.util.fixture.QueuePaymentEventFixture.aQueuePaymentEventFixture;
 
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -50,6 +51,63 @@ public void shouldInsertTransaction() {
...
+
+        transactionDao.upsert(transactionEntity);
+
+        TransactionEntity retrievedTransaction = transactionDao.findTransactionByExternalId(transactionEntity.getExternalId()).get();
+
+        assertThat(retrievedTransaction.getSource(), is(Source.CARD_API));
+    }
+
+    @Test
+    public void shouldInsertTransactionWithSourceCardPaymentLink() {
```
kbottla: This feels redundant. Don't think we need to have a test for each Source constant for integration test. Could cover possible combinations in unit tests

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -50,6 +51,63 @@ public void shouldInsertTransaction() {
...
+
+        transactionDao.upsert(transactionEntity);
+
+        TransactionEntity retrievedTransaction = transactionDao.findTransactionByExternalId(transactionEntity.getExternalId()).get();
+
+        assertThat(retrievedTransaction.getSource(), is(Source.CARD_PAYMENT_LINK));
+    }
+
+    @Test
+    public void shouldInsertTransactionWithSourceCardExternalTelephone() {
```
kbottla: Same as above.

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
rorymalcolm: agree

src/test/java/uk/gov/pay/ledger/event/model/TransactionEntityFactoryTest.java
```diff
@@ -71,7 +72,6 @@ public void fromShouldConvertEventDigestToTransactionEntity() {
         assertThat(transactionEntity.getTransactionType(), is("PAYMENT"));
         assertThat(transactionEntity.isLive(), is(true));
 
-
```
kbottla: Don't we need assertion removed for not null source ? `assertThat(transactionEntity.getSource().toString(), is("CARD_API"));`

PR #525, opened 2020-01-16
Merge pull request #525 from alphagov/PP-6013-add-test-for-source-enum-db-type

rorymalcolm:
- There is a type in the database which contains all the values of the
current enum Source from pay-java-commons, if a mismatch between this
types were to occur this would be problematic

- This test will catch any mismatches between the types in the database
and the types in the Source enum

- Also includes a reformat to match code conventions regarding DB query string styling across ledger

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
@@ -10,13 +10,18 @@ import uk.gov.pay.ledger.transaction.model.TransactionType;
 import uk.gov.pay.ledger.transaction.state.TransactionState;
 import uk.gov.pay.ledger.util.fixture.TransactionFixture;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -280,4 +285,10 @@ public void findTransactionByParentIdAndGatewayAccountId_shouldFilterByParentExt
         assertThat(transactionEntity.getId(), is(transactionWithParentExternalId.getId()));
 
     }
+
+    @Test
+    public void sourceTypeInDatabase_shouldMatchValuesInEnum() {
+        var sourceArray = Arrays.stream(Source.values()).map(Enum::toString).collect(Collectors.toList());
+        transactionDao.getSourceTypeValues().forEach(x -> assertThat(sourceArray.contains(x), is(true)));
```
kbottla: Should this assertion be other way around ? i.e., for all Source enum constants there should be an entry in DB

If a new enum constant say `CARD_IVR` is added to enum but not to DB, then inserting a transaction record with source CARD_IVR will result in Postgres exception

PR #535, opened 2020-01-21
PP-6029 Add java buildpack manifest

danworth:
Adds a java buildpack manifest and dev.yml with common development
environment variables. To deploy into a PaaS space log into the space
and run the following from this project's root::

```
cf push --vars-file dev.yml \
  --var space=<space> \
  --var ledger_route=ledger-<space>.apps.internal

```

**How to test**
Push to your dev space. If its the first time you may need to delete the existing docker image based app with `cf delete ledger` and then recreate the postgres network policy with
`cf add-network-policy ledger --destination-app postgres --protocol tcp --port 5432` (and the same for sqs) once the new buildpack based version of ledger exists (likely it'll fail to start first time because it won't be able to connect to postgres).

### DIFF
--- a/dev.yml
+++ b/dev.yml
@@ -1,0 +1,12 @@
+memory: 500M
+db_password: mysecretpassword
+db_user: ledger
+db_name: ledger
+db_ssl_option: ssl=none
+disk_quota: 500M
+disable_internal_https: 'true'
+run_migration: 'true'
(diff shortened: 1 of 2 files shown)


danworth:
After our discussion I think we've agreed to keep the DB_HOST etc as-is.

PR #540, opened 2020-01-22
PP-6010 - Adds TransactionIT for searching refunds with display size two

nimalank7:
Description:
- This adds an integration test to partially replace SearchRefundsIT.shouldSearchForAllExistingRefundsWithDisplaySizeTwo in 'publicapi' pay-endtoend
- A pact test publicapi-ledger-search-refunds-display-size-two.json in publicapi to cover the remaining functionality

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
@@ -333,6 +333,45 @@ public class TransactionResourceIT {
     }
 
     @Test
+    public void shouldSearchForAllExistingRefundsWithDisplaySizeTwo() {
+        String gatewayAccountId = "2";
+        TransactionFixture createdTransaction = aTransactionFixture()
+                .withTransactionType("REFUND")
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
```diff
@@ -332,6 +332,45 @@ public void shouldSearchCorrectlyUsingRefundStates() {
                 .body("results[1].transaction_id", is(submittedTransaction.getExternalId()));
     }
 
+    @Test
+    public void shouldSearchForAllExistingRefundsWithDisplaySizeTwo() {
```
stephencdaly: The test looks good, but looks like searching with a display size is already covered by `shouldSearchUsingAllFieldsAndReturnAllFieldsCorrectly` so I don't think we should add it. Reason being that integration tests are relatively expensive to run, so we should have the minimum possible set to cover functionality.

nimalank7: Sure I'll close the PR

PR #541, opened 2020-01-23
PP-6059 Create metadata tables

kbottla:
## WHAT

- Creates new tables `metadata_key`, `transaction_metadata` to store to external metadata keys for transactions.
- metadata `key` is set to `50` characters long, which is over the limit `30` we allow services

### DIFF
--- a/src/main/resources/migrations/00033_create_table_metadata_key.sql
+++ b/src/main/resources/migrations/00033_create_table_metadata_key.sql
@@ -1,0 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_metadata_key
+CREATE TABLE metadata_key (
+    id BIGSERIAL PRIMARY KEY,
+    key VARCHAR(50) NOT NULL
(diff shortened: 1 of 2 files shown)


src/main/resources/migrations/00033_create_table_metadata_key.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_metadata_key
+CREATE TABLE metadata_key (
+    id BIGSERIAL PRIMARY KEY,
+    key VARCHAR(50) NOT NULL
```
sfount: Is there anywhere we enforce the 50 character limit anywhere else? I'm guessing nothing will ever get this long but just in case!

kbottla: I think both connector and publicapi enforce it to max of 30 characters. If we ever change these limits, ledger should be updated too

PR #546, opened 2020-01-24
PP-6059 Add TransactionMetadataDao

kbottla:
## WHAT
- Added TransactionMetadataDao with a method to record transaction and metadatakey relation
- Relevant tests
- Correct migration script file name (DB hasn't been migrated yet in any environment)
   00035_table_transaction_external_metadata.sql → 00035_partial_index_transaction_external_metadata.sql


src/main/java/uk/gov/pay/ledger/transactionmetadata/dao/TransactionMetadataDao.java
```diff
@@ -0,0 +1,21 @@
...
+import org.jdbi.v3.sqlobject.customizer.Bind;
+import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
+import org.jdbi.v3.sqlobject.statement.SqlUpdate;
+
+import java.util.Optional;
+
+public interface TransactionMetadataDao {
+
+    @SqlUpdate("INSERT INTO transaction_metadata(transaction_id, metadata_key_id) " +
+            "SELECT :transactionId, (select id from metadata_key where key = :key) " +
```
sfount: Does this duplicated sub query (`(select id from metadata_key where key = :key)`) get processed twice or is postgres smart enough to use the same result? 

I wonder if checking for the existence of all keys in one query and then trusting they should be inserted in another would change the performance; this looks good to me though.

PR #551, opened 2020-01-27
PP-6058 Index id, created_date on transaction table

sfount:
In order to use keyset pagination through transactions on asynchronous
(CSV) downloads, add compound index across both id and transaction
created_date.

This allows ordering by created_date and then id, providing a guaranteed
unique and consistent page without overlapping entities. Created date on
its own could is likely to contain duplicates and we cannot gaurantee
the order of transaction ids as events are eventually consistent.

### DIFF
--- a/src/main/resources/migrations/00033_compound_index_transaction_id_created_date.sql
+++ b/src/main/resources/migrations/00033_compound_index_transaction_id_created_date.sql
@@ -1,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:compound_index_transaction_created_date_id runInTransaction:false
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS transaction_created_date_id_idx ON transaction USING btree(created_date, id);


src/main/resources/migrations/00033_compound_index_transaction_id_created_date.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:compound_index_transaction_id_created_date runInTransaction:false
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS id_created_date_idx ON transaction USING btree(id, created_date);
```
kbottla: Should the columns be other way around `created_date, id` as we do order by `created_date, id`

kbottla: Can the index be named like `transaction_....._idx` as we do for rest of the indexes on transaction table.

sfount: Good catch!

PR #554, opened 2020-01-28
PP-6060 add TransactionMetadataService

SandorArpa:
## WHAT

- add a new service to store metadata for transactions
- service reads metadata from Event
- add relevant tests


src/test/java/uk/gov/pay/ledger/queue/EventMessageHandlerTest.java
```diff
@@ -31,6 +32,9 @@
     @Mock
     private TransactionService transactionService;
 
+    @Mock
+    private TransactionMetadataService transactionMetadataService;
```
kbottla: This is not used

src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
```diff
@@ -61,6 +65,7 @@ private void processSingleMessage(EventMessage message) throws QueueException {
         if(response.isSuccessful()) {
             EventDigest eventDigest = eventService.getEventDigestForResource(event.getResourceExternalId());
             transactionService.upsertTransactionFor(eventDigest);
+            transactionMetadataService.upsertMetadataFor(eventDigest);
```
kbottla: Don't think we want to process on `eventDigest`. 

Event digest includes all events for a given transaction. So with this, we will be processing metadata every time we get event for a transaction and could impact performance. We could only process metadata if event payload has `external_metadata` object

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
```diff
@@ -0,0 +1,45 @@
...
+                        metadataKeyDao.insertIfNotExist(metadataKey);
+                        transactionMetadataDao
+                                .insertIfNotExist(transactionEntity.getId(), metadataKey);
+                    });
+                });
+    }
+
+    private List<String> metadataKeyListFrom(EventDigest eventDigest) {
+        Map<String, Object> metadata =
+                (Map<String, Object>) eventDigest.getEventPayload().get("external_metadata");
```
kbottla: Not sure if nested object is returned as Map<String,Object> too with https://github.com/alphagov/pay-ledger/blob/master/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java#L83.  Would be good to have a test if not exists and if possible.

src/test/java/uk/gov/pay/ledger/queue/EventMessageHandlerTest.java
SandorArpa: Neither is TransactionService, but without either mocks the test fails:
`Wanted but not invoked:
eventQueue.markMessageAsProcessed(
    <any uk.gov.pay.ledger.queue.EventMessage>
);`

SandorArpa: Neither is TransactionService, but without either mocks the test fails:
Wanted but not invoked: eventQueue.markMessageAsProcessed( <any uk.gov.pay.ledger.queue.EventMessage> );

kbottla: 👍 didn't notice `@InjectMocks`

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
```diff
@@ -0,0 +1,52 @@
...
+        this.transactionDao = transactionDao;
+    }
+
+    public void upsertMetadataFor(Event event) {
+        JsonNode eventDataNode;
+        try {
+            eventDataNode = new ObjectMapper().readTree(event.getEventData());
+        } catch (JsonProcessingException e) {
+            LOGGER.error("Unable to parse incoming event payload: {}", e.getMessage());
+            return;
```
kbottla: Can we throw exception if unable to parse event payload ? If not, there is a chance we won't have visibility for errors.

if exception is thrown, events will be placed in dead letter queue for which we will be alerted

kbottla: Given that exception is thrown https://github.com/alphagov/pay-ledger/blob/66b7312a20e0046d3d33c9e9f6c5c5dda9256f7c/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java#L85 for invalid JSON we might not need to throw exceptions in later stages.

PR #557, opened 2020-01-29
PP-6061 Add findMetadataKeysForTransactions DAO method

kbottla:
## WHAT 
- Added new method `findMetadataKeysForTransactions` to derive metadata keys from new tables (transaction_metadata and metadata_key) for given search parameters
- Changed `TransactionMetadataDao` to a class to add concrete method
- Relevant test

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transactionmetadata/dao/TransactionMetadataDao.java
+++ b/src/main/java/uk/gov/pay/ledger/transactionmetadata/dao/TransactionMetadataDao.java
@@ -1,21 +1,74 @@
 package uk.gov.pay.ledger.transactionmetadata.dao;
 
+import com.google.inject.Inject;
+import org.apache.commons.lang3.StringUtils;
+import org.jdbi.v3.core.Jdbi;
+import org.jdbi.v3.core.statement.Query;
 import org.jdbi.v3.sqlobject.customizer.Bind;
-import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transactionmetadata/dao/TransactionMetadataDao.java
```diff
@@ -1,21 +1,74 @@
...
+            "      and transaction_id = :transactionId)";
+
+    private final Jdbi jdbi;
+
+    @Inject
+    public TransactionMetadataDao(Jdbi jdbi) {
+        this.jdbi = jdbi;
+    }
+
+    public void insertIfNotExist(@Bind("transactionId") Long transactionId,
```
sfount: We shouldn't need to `@Bind` in the params now that we're manually doing it on jdbi

EDIT: beat me to it

src/main/java/uk/gov/pay/ledger/transactionmetadata/dao/TransactionMetadataDao.java
```diff
@@ -1,21 +1,74 @@
...
+            Query query = handle.createQuery(queryString);
+            searchParams.getQueryMap().forEach(bindSearchParameter(query));
+
+            return query
+                    .mapTo(String.class)
+                    .list();
+        });
+    }
+
+    private BiConsumer<String, Object> bindSearchParameter(Query query) {
```
sfount: It might be worth both the metadata and transaction search dao using the same binding etc. 

Potentially when we come for a second pass of this code.

PR #572, opened 2020-02-05
bau: Consolidate publicapi->ledger search pact

oswaldquek:
This fixes https://github.com/alphagov/pay-publicapi/pull/855

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/pact/ContractTest.java
+++ b/src/test/java/uk/gov/pay/ledger/pact/ContractTest.java
@@ -44,6 +44,29 @@ public abstract class ContractTest {
         helper.truncateAllData();
     }
 
+    @State("a payment transaction exists")
+    public void aPaymentTransactionExists(Map<String, String> params) {
+        String transactionExternalId = params.get("transaction_external_id");
+        String gatewayAccountId = params.get("gateway_account_id");
+        String cardholderName = params.get("cardholder_name");
+
+        aTransactionFixture()
+                .withExternalId(transactionExternalId)
+                .withGatewayAccountId(gatewayAccountId)
+                .withAmount(1000L)
+                .withReference("aReference")
+                .withDescription("Test description")
+                .withState(TransactionState.CREATED)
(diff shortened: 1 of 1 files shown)


src/test/java/uk/gov/pay/ledger/pact/ContractTest.java
```diff
@@ -44,6 +44,29 @@ public void setUp() {
         helper.truncateAllData();
     }
 
+    @State("a payment transaction exists")
+    public void aPaymentTransactionExists(Map<String, String> params) {
+        String transactionExternalId = params.get("transaction_external_id");
+        String gatewayAccountId = params.get("gateway_account_id");
+        String cardholderName = params.get("cardholder_name");
```
stephencdaly: is the only difference that this state takes a cardholder name param? could the state below not check if that param has been given and add it to the transaction if so?

oswaldquek: it could but i wanted to avoid potential conflicts with other tests that are using the same state. having searched, there's one other test that uses the state below. the state below will eventually be deleted though.

stephencdaly: if the param is not there, it could just have the same behaviour as it already has, right?

stephencdaly: saves having to update twice right? I'll approve though and you can do whatever's easier

oswaldquek: hmm well i want to do things right. https://github.com/alphagov/pay-publicapi/pull/855 adds this to the pact:
```
        {
          "name": "a payment transaction exists",
          "params": {
            "transaction_external_id": "charge97837509646393e3C",
            "gateway_account_id": "123456",
            "cardholder_name": "j.doe@example.org"
          }
        }
```
so i want to do things properly rather than be too clever.

stephencdaly: ah didn't see you already had a pr open expecting this state

PR #585, opened 2020-02-10
PP-6139 View MOTO Transactions per gateway

nimalank7:
Description:
- Adds moto field to Ledger endpoint so that when moto flag is passed it will return the count. This is so that it can be displayed in the statistics page for the individual gateway account in Toolbox
- Since moto payments isn't a transaction type in Ledger a new method in the DAO and service layer was created to just return moto payments
- Follow up PR will address tests for DAO, service layer and validation


src/main/java/uk/gov/pay/ledger/report/params/TransactionSummaryParams.java
```diff
@@ -68,12 +81,27 @@ public void setToDate(String toDate) {
...
     public Map<String, Object> getQueryMap() {
         Map<String, Object> queryMap = new HashMap<>();
 
         if (isNotBlank(accountId)) {
             queryMap.put(GATEWAY_ACCOUNT_EXTERNAL_FIELD, accountId);
         }
+
+        if (moto) {
+            queryMap.put(MOTO, true);
+        }
```
stephencdaly: I think this is getting a bit confused between the query parameter indicating whether or not we should include statistics about the moto payments and whether payments are moto or not. Might be easier to talk it through.

src/main/java/uk/gov/pay/ledger/report/params/TransactionSummaryParams.java
```diff
@@ -13,13 +13,13 @@
...
 
 public class TransactionSummaryParams {
-    private static final String TO_DATE_FIELD = "to_date";
-    private static final String GATEWAY_ACCOUNT_EXTERNAL_FIELD = "account_id";
-    private static final String FROM_DATE_FIELD = "from_date";
 
     @QueryParam("account_id")
     private String accountId;
 
+    @QueryParam("moto")
```
stephencdaly: Maybe this name could be slightly more representative of what it does. Possibly along the lines of `include_moto_statistics`

src/test/java/uk/gov/pay/ledger/report/dao/builder/TransactionStatisticQueryTest.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.junit.Before;
+import org.junit.Test;
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasSize;
+import static org.hamcrest.Matchers.is;
+
+public class TransactionStatisticQueryTest {
```
stephencdaly: would be good to add tests for adding the moto constraint to the query

src/test/java/uk/gov/pay/ledger/report/service/ReportServiceTest.java
```diff
@@ -36,32 +38,33 @@
...
 
     @Captor
-    private ArgumentCaptor<TransactionSummaryParams> transactionSummaryParamsArgumentCaptor;
+    private ArgumentCaptor<TransactionStatisticQuery> transactionStatisticQueryArgumentCaptor;
 
     @InjectMocks
     private ReportService reportService;
 
     @Test
     public void shouldReturnPaymentCountsByState_whenGatewayAccountIdProvided() {
```
stephencdaly: should probably add in service tests for whether or not we're including moto statistics too

src/main/java/uk/gov/pay/ledger/report/service/ReportService.java
```diff
@@ -29,19 +33,49 @@ public ReportService(ReportDao reportDao) {
...
-        TransactionsStatisticsResult refunds = reportDao.getTransactionSummaryStatistics(params, TransactionType.REFUND);
-        return new TransactionSummaryResult(payments, refunds, payments.getGrossAmount() - refunds.getGrossAmount());
+        TransactionStatisticQuery transactionStatisticQuery = buildBaseTransactionStatisticQuery(params);
+
+        TransactionsStatisticsResult payments = reportDao.getTransactionSummaryStatistics(transactionStatisticQuery, TransactionType.PAYMENT);
+        TransactionsStatisticsResult refunds = reportDao.getTransactionSummaryStatistics(transactionStatisticQuery, TransactionType.REFUND);
+        TransactionsStatisticsResult motoPayments = new TransactionsStatisticsResult(0L, 0L);
+
+        if (params.isIncludeMotoStatistics()) {
+            TransactionStatisticQuery motoQuery = transactionStatisticQuery.withMoto(params.isIncludeMotoStatistics());
```
stephencdaly: I think that this should either be `transactionStatisticQuery.withMoto(true)` or possibly `withMoto` shouldn't take parameters. It's exactly the same in practice, but maybe slightly confusing to read as it is.

PR #589, opened 2020-02-12
PP-6076 Include MOTO column in CSV

stephencdaly:
If the GET /v1/transaction endpoint is called with Accept header "text/csv" and query parameter `moto_header=true` then include a MOTO column with a boolean type as the final column before the metadata column.

If the query parameter is not provided or false then do not include the column.

Add some additional tests to TransactionResourceIT to test for the including/excluding the optional columns, including the "Net" and "Fee" columns. Inline a private method in this test class that was only used in one place.


src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
```diff
@@ -717,7 +717,107 @@ public void shouldGetAllTransactionsAsCSVWithAcceptType() throws IOException {
                 .contentType("text/csv")
                 .extract().asInputStream();
 
-        assertHealthyCsvResponse(csvResponseStream, transactionFixture);
+        List<CSVRecord> csvRecords = CSVParser.parse(csvResponseStream, Charset.defaultCharset(), RFC4180.withFirstRecordAsHeader()).getRecords();
```
alexbishop1: I know you just copied this from what was already there but `Charset.defaultCharset()` almost certainly wants to be `StandardCharsets.UTF_8` (since that’s what we should be outputting from the transaction resource).

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
```diff
@@ -790,41 +901,11 @@ public void shouldReturnCSVHeadersInCorrectOrder() throws IOException {
...
-        assertThat(refundRecord.get("Time Created"), is("16:24:01"));
-        assertThat(refundRecord.get("Corporate Card Surcharge"), is("0.00"));
-        assertThat(refundRecord.get("Total Amount"), is("-1.00"));
-        assertThat(refundRecord.get("Wallet Type"), is(""));
-        assertThat(refundRecord.get("Issued By"), is("refund-by-user-email@example.org"));
+        assertThat(header.get(20), is("Fee"));
+        assertThat(header.get(21), is("Net"));
+        assertThat(header.get(22), is("Card Type"));
+        assertThat(header.get(23), is("MOTO"));
+        assertThat(header.get(24), is("a-metadata-key (metadata)"));
```
alexbishop1: This class is almost 1,000 lines long — might it be worth extracting the CSV tests into their own file? Some of the inserting of transactions that happens in each CSV test could probably be in to the `@Before` method then.

src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
```diff
@@ -154,7 +156,9 @@ public CsvTransactionFactory(ObjectMapper objectMapper) {
         return result;
     }
 
-    public Map<String, Object> getCsvHeadersWithMedataKeys(List<String> metadataKeys, boolean includeFeeHeaders) {
+    public Map<String, Object> getCsvHeadersWithMedataKeys(List<String> metadataKeys,
+                                                           boolean includeFeeHeaders,
+                                                           boolean includeMotoHeader) {
```
alexbishop1: Adding another boolean is fine here but I think if we find ourselves adding any more we should refactor to use an `EnumSet` or something to represent which headers we want to include

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
stephencdaly: cool, will change

src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
stephencdaly: Yeh I agree

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
stephencdaly: Yeh good idea, it was difficult to find the relevant tests. I'll add another commit to move tests out.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
stephencdaly: done

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
stephencdaly: Split out the CSV and search tests. Didn't move any of the transaction setup to the `@Before` for CSV as the tests do need slightly different transactions

PR #607, opened 2020-02-20
PP-6163 Search For Multiple Gateway Accounts

rorymalcolm:
- Ledger now allows users to supply comma separated lists of account IDs
via the `account_id` parameter, this will then return payments for all
of the gateway accounts specified.

- Adds tests to test the functionality works as expected.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
@@ -46,7 +46,7 @@ public class TransactionSearchParams {
     @DefaultValue("false")
     @QueryParam("with_parent_transaction")
     private boolean withParentTransaction;
-    private String accountId;
+    private List<String> accountIds;
     @QueryParam("email")
     private String email;
     @QueryParam("reference")
@@ -80,8 +80,8 @@ public class TransactionSearchParams {
     private Map<String, Object> queryMap;
     private String gatewayTransactionId;
 
-    public void setAccountId(String accountId) {
-        this.accountId = accountId;
+    public void setAccountIds(List<String> accountIds) {
+        this.accountIds = List.copyOf(accountIds);
     }
 
     public void setEmail(String email) {
@@ -221,8 +221,8 @@ public class TransactionSearchParams {
 
     private void addCommonFilterTemplates(List<String> filters) {
 
-        if (isNotBlank(accountId)) {
-            filters.add(" t.gateway_account_id = :" + GATEWAY_ACCOUNT_EXTERNAL_FIELD);
+        if (accountIds != null && !accountIds.isEmpty()) {
+            filters.add(" t.gateway_account_id IN (<" + GATEWAY_ACCOUNT_EXTERNAL_FIELD + ">)");
         }
         if (transactionType != null) {
             filters.add(" t.type = :" + TRANSACTION_TYPE_FIELD + "::transaction_type");
@@ -249,8 +249,8 @@ public class TransactionSearchParams {
         if (queryMap == null) {
             queryMap = new HashMap<>();
 
-            if (isNotBlank(accountId)) {
-                queryMap.put(GATEWAY_ACCOUNT_EXTERNAL_FIELD, accountId);
+            if (accountIds != null && !accountIds.isEmpty()) {
+                queryMap.put(GATEWAY_ACCOUNT_EXTERNAL_FIELD, accountIds);
             }
 
             if (isNotBlank(email)) {
@@ -309,8 +309,8 @@ public class TransactionSearchParams {
                 TransactionState.getStatesForOldStatus(status);
     }
 
-    public String getAccountId() {
-        return accountId;
+    public List<String> getAccountIds() {
+        return accountIds;
     }
 
     public Long getPageNumber() {
@@ -348,8 +348,8 @@ public class TransactionSearchParams {
     public String buildQueryParamString(Long forPage) {
         List<String> queries = new ArrayList<>();
 
-        if (isNotBlank(accountId)) {
-            queries.add(GATEWAY_ACCOUNT_EXTERNAL_FIELD + "=" + accountId);
+        if (accountIds != null && !accountIds.isEmpty()) {
+            queries.add(GATEWAY_ACCOUNT_EXTERNAL_FIELD + "=" + String.join(",", accountIds));
         }
         if (isNotBlank(fromDate)) {
             queries.add(FROM_DATE_FIELD + "=" + fromDate);
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
@@ -6,22 +6,20 @@ import uk.gov.pay.ledger.exception.ValidationException;
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
 
-import static org.apache.commons.lang3.StringUtils.isBlank;
 import static org.apache.commons.lang3.StringUtils.isNotBlank;
 
 public class TransactionSearchParamsValidator {
     private static final String FROM_DATE_FIELD = "from_date";
     private static final String TO_DATE_FIELD = "to_date";
 
-    public static void validateSearchParams(TransactionSearchParams searchParams, String gatewayAccountId) {
+    public static void validateSearchParams(TransactionSearchParams searchParams, CommaDelimitedSetParameter gatewayAccountIds) {
         if (isNotBlank(searchParams.getFromDate())) {
             validateDate(FROM_DATE_FIELD, searchParams.getFromDate());
         }
         if (isNotBlank(searchParams.getToDate())) {
             validateDate(TO_DATE_FIELD, searchParams.getToDate());
         }
-
-        if(searchParams.getWithParentTransaction() && isBlank(gatewayAccountId)){
+        if(searchParams.getWithParentTransaction() && gatewayAccountIds.getParameters().isEmpty()){
             throw new ValidationException("gateway_account_id is mandatory to search with parent transaction");
         }
     }
--- a/src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
@@ -1,0 +1,52 @@
+package uk.gov.pay.ledger.transaction.service;
+
+import uk.gov.pay.ledger.exception.ValidationException;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static java.lang.String.format;
+
+public class AccountIdListSupplierManager<T> {
+
+    private static final String ACCOUNT_ID = "account_id";
+
+    private final boolean overrideAccountRestriction;
+    private final List<String> gatewayAccountIds;
+
+    private Supplier<T> privilegedSupplier;
+    private Function<List<String>, T> supplier;
+
+    public AccountIdListSupplierManager(Boolean overrideAccountRestriction, List<String> gatewayAccountIds) {
+        this.overrideAccountRestriction = Optional.ofNullable(overrideAccountRestriction).orElse(false);
+        this.gatewayAccountIds = gatewayAccountIds != null ? gatewayAccountIds : List.of();
+    }
+
+    public static <U> AccountIdListSupplierManager<U> of(Boolean overrideAccountRestriction, List<String> gatewayAccountIds) {
+        return new AccountIdListSupplierManager<>(overrideAccountRestriction, gatewayAccountIds);
+    }
+
+    public AccountIdListSupplierManager<T> withPrivilegedSupplier(Supplier<T> supplier) {
+        this.privilegedSupplier = supplier;
+        return this;
+    }
+
+    public AccountIdListSupplierManager<T> withSupplier(Function<List<String>, T> supplier) {
+        this.supplier = supplier;
+        return this;
+    }
+
+    public T validateAndGet() {
+        if (!overrideAccountRestriction && gatewayAccountIds.isEmpty()) {
+            throw new ValidationException(format("Field [%s] cannot be empty", ACCOUNT_ID));
+        }
+
+        if (gatewayAccountIds.isEmpty()) {
+            return privilegedSupplier.get();
+        }
+
+        return supplier.apply(gatewayAccountIds);
+    }
+}
--- a/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
@@ -2,7 +2,6 @@ package uk.gov.pay.ledger.transaction.service;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
-import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.ledger.event.dao.EventDao;
@@ -85,12 +84,13 @@ public class TransactionService {
     }
 
     public TransactionSearchResponse searchTransactions(TransactionSearchParams searchParams, UriInfo uriInfo) {
-        return searchTransactions(null, searchParams, uriInfo);
+        return searchTransactions(List.of(), searchParams, uriInfo);
     }
 
-    public TransactionSearchResponse searchTransactions(String gatewayAccountId, TransactionSearchParams searchParams, UriInfo uriInfo) {
-        if (StringUtils.isNotBlank(gatewayAccountId)) {
-            searchParams.setAccountId(gatewayAccountId);
+    public TransactionSearchResponse searchTransactions(List<String> gatewayAccountIds,
+                                                        TransactionSearchParams searchParams, UriInfo uriInfo) {
+        if (!gatewayAccountIds.isEmpty()) {
+            searchParams.setAccountIds(gatewayAccountIds);
         }
 
         if (searchParams.getWithParentTransaction()) {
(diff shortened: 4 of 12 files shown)


src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
```diff
@@ -6,22 +6,21 @@
...
+    public static void validateSearchParams(TransactionSearchParams searchParams, CommaDelimitedSetParameter gatewayAccountIds) {
         if (isNotBlank(searchParams.getFromDate())) {
             validateDate(FROM_DATE_FIELD, searchParams.getFromDate());
         }
         if (isNotBlank(searchParams.getToDate())) {
             validateDate(TO_DATE_FIELD, searchParams.getToDate());
         }
-
-        if(searchParams.getWithParentTransaction() && isBlank(gatewayAccountId)){
+        if(searchParams.getWithParentTransaction() && isEmpty(gatewayAccountIds.getParameters())){
```
alexbishop1: You can just do `gatewayAccountIds.getParameters().isEmpty()` and avoid the need for Guava.

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+import static java.lang.String.format;
+import static org.apache.commons.lang3.ObjectUtils.isEmpty;
+
+public class AccountIdListSupplierManager<T> {
+
+    private static final String GATEWAY_ACCOUNT_ID = "account_id";
```
alexbishop1: Just `ACCOUNT_ID` would be more accurate.

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private Supplier<T> privilegedSupplier;
+    private Function<List<String>, T> supplier;
+
+    public AccountIdListSupplierManager(Boolean overrideAccountRestriction, List<String> gatewayAccountIds) {
+        this.overrideAccountRestriction = Optional.ofNullable(overrideAccountRestriction).orElse(false);
+        this.gatewayAccountIds = gatewayAccountIds;
+    }
+
+    public static <U> AccountIdListSupplierManager<U> of(Boolean overrideAccountRestriction, List<String> gatewayAccountId) {
```
alexbishop1: Probably want `gatewayAccountId` to be `gatewayAccountIds`.

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
```diff
@@ -0,0 +1,53 @@
...
+        return this;
+    }
+
+    public AccountIdListSupplierManager<T> withSupplier(Function<List<String>, T> supplier) {
+        this.supplier = supplier;
+        return this;
+    }
+
+    public T validateAndGet() {
+        if(!overrideAccountRestriction && isEmpty(gatewayAccountIds)) {
```
alexbishop1: Did you really mean to use `org.apache.commons.lang3.ObjectUtils.isEmpty` here? If you want the `null` check, I would just do…

```java
if (!overrideAccountRestriction && (gatewayAccountIds == null || gatewayAccountIds.isEmpty())) {
```

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -88,9 +89,10 @@ public TransactionSearchResponse searchTransactions(TransactionSearchParams sear
         return searchTransactions(null, searchParams, uriInfo);
     }
 
-    public TransactionSearchResponse searchTransactions(String gatewayAccountId, TransactionSearchParams searchParams, UriInfo uriInfo) {
-        if (StringUtils.isNotBlank(gatewayAccountId)) {
-            searchParams.setAccountId(gatewayAccountId);
+    public TransactionSearchResponse searchTransactions(List<String> gatewayAccountIds,
+                                                        TransactionSearchParams searchParams, UriInfo uriInfo) {
+        if (!isEmpty(gatewayAccountIds)) {
```
alexbishop1: If the callers of this were just changed so that they pass an empty list rather than `null`, this could just be `gatewayAccountIds.isEmpty()`.

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoSearchIT.java
```diff
@@ -843,13 +843,46 @@ public void searchTransactionsByCursor() {
...
+        aTransactionFixture()
+                .withGatewayAccountId("1337")
+                .withCreatedDate(now(ZoneOffset.UTC).minusDays(1))
+                .insert(rule.getJdbi());
+
+        searchParams.setAccountIds(List.of("1","2"));
+
+        List<TransactionEntity> transactionList = transactionDao.searchTransactions(searchParams);
+
+        assertThat(transactionList.size(), is(4));
```
alexbishop1: Might be worth sticking some sort of unique identifier (e.g. an external ID) on each transaction so we can verify we’re getting the right ones.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceSearchIT.java
```diff
@@ -408,4 +408,73 @@ public void shouldSearchCorrectlyOnTransactionAndParentTransaction_WhenWithParen
                 .body("results[1].transaction_id", is(payment.getExternalId()))
                 .body("results[1].parent_transaction", is(nullValue()));
     }
+
+    @Test
+    public void shouldSearchUsingMultipleGatewayAccountIds() {
```
alexbishop1: I think this test is redundant with the below one.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceSearchIT.java
```diff
@@ -408,4 +408,73 @@ public void shouldSearchCorrectlyOnTransactionAndParentTransaction_WhenWithParen
...
+                        "&display_size=5"
+                )
+                .then()
+                .statusCode(Response.Status.OK.getStatusCode())
+                .contentType(JSON)
+                .body("count", is(2));
+    }
+
+    @Test
+    public void shouldSearchUsingMultipleGatewayAccountIdsExclusively() {
```
alexbishop1: What’s “exclusively” mean in this case?

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceSearchIT.java
```diff
@@ -408,4 +408,73 @@ public void shouldSearchCorrectlyOnTransactionAndParentTransaction_WhenWithParen
...
+        aTransactionFixture()
+                .withTransactionType("PAYMENT")
+                .withState(TransactionState.SUBMITTED)
+                .withGatewayAccountId(otherGatewayAccountId)
+                .insert(rule.getJdbi());
+
+        aTransactionFixture()
+                .withTransactionType("PAYMENT")
+                .withState(TransactionState.SUBMITTED)
+                .withGatewayAccountId(otherestGatewayAccountId)
```
alexbishop1: Not convinced “otherest” is a word but I’ll let it pass.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceSearchIT.java
```diff
@@ -408,4 +408,73 @@ public void shouldSearchCorrectlyOnTransactionAndParentTransaction_WhenWithParen
...
+                .insert(rule.getJdbi());
+
+        given().port(port)
+                .contentType(JSON)
+                .accept(JSON)
+                .get("/v1/transaction?" +
+                        "account_id=" + targetGatewayAccountId+ "," + otherGatewayAccountId+
```
alexbishop1: Looks like the `otherGatewayAccountId` is also a `targetGatewayAccountId`, really.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
rorymalcolm: not if it's null

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
rorymalcolm: does `gatewayAccountIds == null || gatewayAccountIds.isEmpty()` not throw an NPE?

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceSearchIT.java
rorymalcolm: accounts that aren't searched for dont have their transactions returned

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
alexbishop1: If `gatewayAccountIds == null` it won’t evaluate the other side because it already knows that the whole logical or expression will return `true`.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -221,8 +222,8 @@ public void overrideMaxDisplaySize(Long maxDisplaySize) {
 
     private void addCommonFilterTemplates(List<String> filters) {
 
-        if (isNotBlank(accountId)) {
-            filters.add(" t.gateway_account_id = :" + GATEWAY_ACCOUNT_EXTERNAL_FIELD);
+        if (!isEmpty(accountIds)) {
```
alexbishop1: I still think we can just do `accountIds.isEmpty()`: `setAccountIds(…)` will throw a NPE if `accountIds` is null (because of the `List.copyOf(…)`) and the only caller of that method is in `TransactionResource`’s `streamCsv(…)` method, which will throw its own NPE if the parameter isn’t there. So by this point I’m fairly certain we can be sure that `accountIds` isn’t null.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidator.java
```diff
@@ -6,22 +6,21 @@
 import java.time.ZonedDateTime;
 import java.time.format.DateTimeParseException;
 
-import static org.apache.commons.lang3.StringUtils.isBlank;
+import static com.google.common.collect.Iterables.isEmpty;
```
alexbishop1: Unused import

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
```diff
@@ -0,0 +1,53 @@
...
+        this.supplier = supplier;
+        return this;
+    }
+
+    public T validateAndGet() {
+        if (!overrideAccountRestriction && (gatewayAccountIds == null || gatewayAccountIds.isEmpty())) {
+            throw new ValidationException(format("Field [%s] cannot be empty", ACCOUNT_ID));
+        }
+
+        if(isEmpty(gatewayAccountIds)) {
```
alexbishop1: This can just be `gatewayAccountIds == null || gatewayAccountIds.isEmpty()`.

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -36,6 +36,7 @@
 
 import static java.lang.String.format;
 import static java.util.stream.Collectors.groupingBy;
+import static org.apache.commons.lang3.ObjectUtils.isEmpty;
```
alexbishop1: Unused import

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoSearchIT.java
```diff
@@ -843,13 +843,56 @@ public void searchTransactionsByCursor() {
                 .withCreatedDate(now(ZoneOffset.UTC).minusDays(1))
                 .insert(rule.getJdbi());
 
-        searchParams.setAccountId("1");
+        searchParams.setAccountIds(List.of("1"));
```
alexbishop1: Your country thanks you for your service.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceSearchIT.java
```diff
@@ -408,4 +408,42 @@ public void shouldSearchCorrectlyOnTransactionAndParentTransaction_WhenWithParen
                 .body("results[1].transaction_id", is(payment.getExternalId()))
                 .body("results[1].parent_transaction", is(nullValue()));
     }
+
+    @Test
+    public void shouldSearchUsingMultipleGatewayAccountIdsExclusively() {
```
alexbishop1: I don’t think this needs `Exclusively` in the name now the other test has been removed.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidatorTest.java
```diff
@@ -45,14 +45,14 @@ public void shouldNotThrowException_whenValidDateFormats() {
...
+        TransactionSearchParamsValidator.validateSearchParams(searchParams, new CommaDelimitedSetParameter("account-id"));
     }
 
     @Test
     public void shouldThrowException_whenGatewayAccountIdIsNotAvailableAndWithParentTransactionIsTrue() {
         searchParams.setWithParentTransaction(true);
         thrown.expect(ValidationException.class);
         thrown.expectMessage("gateway_account_id is mandatory to search with parent transaction");
-        TransactionSearchParamsValidator.validateSearchParams(searchParams,"");
+        TransactionSearchParamsValidator.validateSearchParams(searchParams,new CommaDelimitedSetParameter(""));
```
alexbishop1: A test with `new CommaDelimitedSetParameter("1,2,3")` would be good.

src/test/java/uk/gov/pay/ledger/transaction/service/AccountIdSupplierManagerTest.java
```diff
@@ -65,7 +66,7 @@ public void givenAccountIdNotProvidedWhenNotRequiredUsesPrivilegedSupplier() {
     }
 
     private void setUpAndGet(Boolean overrideAccountRestriction, String gatewayAccountId) {
-        new AccountIdSupplierManager(overrideAccountRestriction, gatewayAccountId)
+        new AccountIdListSupplierManager(overrideAccountRestriction, gatewayAccountId == null ? null : List.of(gatewayAccountId))
```
alexbishop1: Doesn’t `AccountIdListSupplierManager` handle `gatewayAccountIds` being null?

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoSearchIT.java
rorymalcolm: lol

src/test/java/uk/gov/pay/ledger/transaction/service/AccountIdSupplierManagerTest.java
rorymalcolm: this is really annoying me, can't work out to fix it atm, it works, something about the mocks, anyway

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidatorTest.java
rorymalcolm: done!!

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
alexbishop1: Whoops, I was wrong here! It’s possible `setAccountIds(…)` is never called at all, so `accountIds` remains at its initial null value.

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
```diff
@@ -0,0 +1,52 @@
...
+        return this;
+    }
+
+    public AccountIdListSupplierManager<T> withSupplier(Function<List<String>, T> supplier) {
+        this.supplier = supplier;
+        return this;
+    }
+
+    public T validateAndGet() {
+        if (!overrideAccountRestriction && (gatewayAccountIds == null || gatewayAccountIds.isEmpty())) {
```
alexbishop1: No need for `gatewayAccountIds == null` check here: the constructor ensures this is never true.

src/main/java/uk/gov/pay/ledger/transaction/service/AccountIdListSupplierManager.java
```diff
@@ -0,0 +1,52 @@
...
+            throw new ValidationException(format("Field [%s] cannot be empty", ACCOUNT_ID));
+        }
+
+        if(gatewayAccountIds == null || gatewayAccountIds.isEmpty()) {
```
alexbishop1: No need for `gatewayAccountIds == null` check here: the constructor ensures this is never true.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsValidatorTest.java
```diff
@@ -45,14 +45,20 @@ public void shouldNotThrowException_whenValidDateFormats() {
...
+    public void shouldNotThrowException_whenGatewayAccountIdsAvailableAndWithParentTransactionIsTrue() {
         searchParams.setWithParentTransaction(true);
         thrown.expect(ValidationException.class);
         thrown.expectMessage("gateway_account_id is mandatory to search with parent transaction");
-        TransactionSearchParamsValidator.validateSearchParams(searchParams,"");
+        TransactionSearchParamsValidator.validateSearchParams(searchParams,new CommaDelimitedSetParameter(""));
+    }
+
+    @Test
+    public void shouldThrowException_whenGatewayAccountIdIsNotAvailableAndWithParentTransactionIsTrue() {
```
alexbishop1: `shouldNotThrowException`, surely?

PR #614, opened 2020-02-25
PP-6186: Add event to indicate status changed to match gateway status

oswaldquek:


oswaldquek:
@stephencdaly oops, good shout. I clearly don't know ledger code well enough. updated.

PR #631, opened 2020-03-09
PP-6201 - Add CAPTURE_NOTIFICATION_FOR_EXPUNGED_CHARGE event to Ledger

nimalank7:
Description:
- This PR corresponds to PP-6201-Force-capture-notifications-for-expunged-charges
- Due to configuration of ePDQ it can be the case that a charge in an terminal state can be captured. Since historic charges aren't in Connector we will send an event to Ledger indicating that it is now captured


src/main/java/uk/gov/pay/ledger/event/model/SalientEventType.java
```diff
@@ -22,6 +22,7 @@
     USER_APPROVED_FOR_CAPTURE_AWAITING_SERVICE_APPROVAL,
     SERVICE_APPROVED_FOR_CAPTURE,
     STATUS_CORRECTED_TO_CAPTURED_TO_MATCH_GATEWAY_STATUS,
+    CAPTURE_NOTIFICATION_FOR_EXPUNGED_CHARGE,
```
sfount: 👋significant events so far have used semantic, past tense naming to describe exactly what happened.

As this maps to an external `SUCCESS` state I'm guessing this means the payment was captured by the gateway? Making this clear will bring this more in line with other event names here. 

Potentially it could be something like: 
`CAPTURE_CONFIRMED_BY_GATEWAT_NOTIFICATION` 

Happy to chat about this if it doesn't line up with the rest of the naming for this feature.

nimalank7: Sure - how about EXPUNGED_CHARGE_STATUS_CORRECTED_TO_CAPTURE_TO_MATCH_GATEWAY_STATUS

src/test/java/uk/gov/pay/ledger/queue/QueueMessageReceiverIT.java
```diff
@@ -155,4 +155,34 @@ public void shouldHandleRefundAvailabilityUpdatedEvent() throws InterruptedExcep
...
+    public void shouldHandleForceCaptureEvent() throws InterruptedException {
+        final String resourceExternalId = "rexid2";
+        final String gatewayAccountId = "test_accountId";
+        aQueuePaymentEventFixture()
+                .withResourceExternalId(resourceExternalId)
+                .withEventDate(CREATED_AT)
+                .withEventType("AUTHORISATION_REJECTED")
+                .withEventData(format("{\"gateway_account_id\":\"%s\"}", gatewayAccountId))
+                .insert(rule.getSqsClient());
+
```
oswaldquek: I think it'll be good to query `"/v1/transaction/" + resourceExternalId + "?account_id=" + gatewayAccountId` here and verify the status is `FAILED_REJECTED` here.
That way when you put the next event type `EXPUNGED_CHARGE_STATUS_CORRECTED_TO_CAPTURED_TO_MATCH_GATEWAY_STATUS` on the queue we are confident the status was actually changed to `success`.

nimalank7: Done

PR #649, opened 2020-03-19
PP-6260 Create Table Payouts

rorymalcolm:
- Create a table `payouts` to store Stripe payout events and include
rollback provisions.

### DIFF
--- a/src/main/resources/migrations/00040_create_table_payouts.sql
+++ b/src/main/resources/migrations/00040_create_table_payouts.sql
@@ -1,0 +1,19 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_payouts
+CREATE TABLE payouts (
+    id BIGSERIAL PRIMARY KEY,
+    gateway_payout_id VARCHAR(26),
+    amount BIGINT NOT NULL,
+    created_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL,
+    paid_out_date TIMESTAMP WITH TIME ZONE,
+    statement_descriptor VARCHAR(255),
+    status VARCHAR(26),
+    type VARCHAR(26),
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00040_create_table_payouts.sql
```diff
@@ -0,0 +1,18 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_payouts
+CREATE TABLE payouts (
+    id BIGSERIAL PRIMARY KEY,
+    external_id VARCHAR(26),
+    amount BIGINT NOT NULL,
+    created_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL,
+    paid_out_date TIMESTAMP WITH TIME ZONE DEFAULT (now() AT TIME ZONE 'utc') NOT NULL,
```
kbottla: not sure default should be now() for paid_out_date. Any chance we don't get it from Stripe

src/main/resources/migrations/00040_create_table_payouts.sql
```diff
@@ -0,0 +1,18 @@
...
+    statement_descriptor VARCHAR(255),
+    status VARCHAR(26),
+    type VARCHAR(26)
```
kbottla: - don't we need `version` ? 
 - is external_id payout id from Stripe ?

src/main/resources/migrations/00040_create_table_payouts.sql
```diff
@@ -0,0 +1,19 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:create_table_payouts
+CREATE TABLE payouts (
+    id BIGSERIAL PRIMARY KEY,
+    gateway_external_id VARCHAR(26),
```
kbottla: anything that has `external_id` may mix up with our terminology. May be gateway_payout_id or something similar ?

PR #655, opened 2020-03-23
PP-6262 Alter Index Migrations

rorymalcolm:
- The index on the payouts table needs to be a unique constraint, this
drops the previous one and adds a new index that the unique constraint
can then be placed against

### DIFF
--- a/src/main/resources/migrations/00041_alter_index_on_payouts_gateway_id.sql
+++ b/src/main/resources/migrations/00041_alter_index_on_payouts_gateway_id.sql
@@ -1,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:drop_index_gateway_payout_id
+DROP INDEX index_gateway_payout_id;
+--rollback CREATE INDEX index_gateway_payout_id ON payouts(gateway_payout_id);
+
+--changeset uk.gov.pay:constraint_index_gateway_payout_id
+ALTER TABLE payouts ADD CONSTRAINT gateway_payout_id_unique  UNIQUE (gateway_payout_id);
+--rollback ALTER TABLE payouts DROP CONSTRAINT gateway_payout_id_unique;


src/main/resources/migrations/00041_alter_index_on_payouts_gateway_id.sql
```diff
@@ -0,0 +1,13 @@
...
+--changeset uk.gov.pay:drop_index_gateway_payout_id
+DROP INDEX index_gateway_payout_id;
+--rollback CREATE INDEX index_gateway_payout_id ON payouts(gateway_payout_id);
+
+--changeset uk.gov.pay:create_unique_index_gateway_payout_id
+CREATE UNIQUE INDEX index_gateway_payout_id ON payouts(gateway_payout_id);
+--rollback DROP INDEX CONCURRENT index_gateway_payout_id;
+
+--changeset uk.gov.pay:constraint_index_gateway_payout_id
+ALTER TABLE payouts ADD CONSTRAINT gateway_payout_id_unique UNIQUE USING INDEX index_gateway_payout_id;
```
kbottla: We may only need unique key constraint and index will be created automatically by postgres https://www.postgresql.org/docs/current/indexes-unique.html
```suggestion
ALTER TABLE payouts ADD CONSTRAINT gateway_payout_id_unique  UNIQUE (gateway_payout_id);
```

src/main/resources/migrations/00041_alter_index_on_payouts_gateway_id.sql
```diff
@@ -0,0 +1,13 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:drop_index_gateway_payout_id
+DROP INDEX index_gateway_payout_id;
+--rollback CREATE INDEX index_gateway_payout_id ON payouts(gateway_payout_id);
+
+--changeset uk.gov.pay:create_unique_index_gateway_payout_id
+CREATE UNIQUE INDEX index_gateway_payout_id ON payouts(gateway_payout_id);
```
kbottla: This is not needed too 

As per docs 
> There's no need to manually create indexes on unique columns; doing so would just duplicate the automatically-created index.

PR #659, opened 2020-03-24
PP-6261 Transaction Table References Payouts

rorymalcolm:
- The `transaction` table now has a foreign key relationship to the
`payouts` table via the `gateway_payout_id`.

### DIFF
--- a/src/main/resources/migrations/00044_add_payouts_foreign_key_to_transaction.sql
+++ b/src/main/resources/migrations/00044_add_payouts_foreign_key_to_transaction.sql
@@ -1,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_payouts_foreign_key_to_transaction
+ALTER TABLE transaction ADD COLUMN gateway_payout_id VARCHAR(50) REFERENCES payouts(gateway_payout_id);
+--rollback ALTER TABLE transaction DROP COLUMN gateway_payout_id; 
+
+--changeset uk.gov.pay:create_index_on_transaction_gateway_payout_id runInTransaction:false
+CREATE INDEX CONCURRENTLY transaction_gateway_payout_id_idx ON transaction(gateway_payout_id);
+--rollback DROP INDEX transaction_gateway_payout_id_idx;


src/main/resources/migrations/00043_add_payouts_foreign_key_to_transaction.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_payouts_foreign_key_to_transaction
+ALTER TABLE transaction ADD COLUMN gateway_payout_id VARCHAR(26) REFERENCES payouts(gateway_payout_id);
```
kbottla: I think we need to increase size of the column as Stripe payout IDs are more than 26 characters ex: `po_1GPf0PJqIiWTNsILGMw50leq`. May be 50 chars or so ? so we don't need to keep changing if Stripe varies length

kbottla: We also need to create index on this column to be able to search for transactions for a given payout but can be a separate PR.

rorymalcolm: ah, we will have to change the length in the payouts table too then

src/main/resources/migrations/00044_add_payouts_foreign_key_to_transaction.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_payouts_foreign_key_to_transaction
+ALTER TABLE transaction ADD COLUMN gateway_payout_id VARCHAR(50) REFERENCES payouts(gateway_payout_id);
+--rollback ALTER TABLE transaction DROP COLUMN gateway_payout_id; 
+
+--changeset uk.gov.pay:create_index_on_transaction_gateway_payout_id
+CREATE INDEX transaction_gateway_payout_id_idx ON transaction(gateway_payout_id);
```
kbottla: Would recommend `create index concurrently` so table won't be locked when building index. Although there is no data in the column it will still lock table for few millis.

PR #677, opened 2020-04-02
Revert "BAU Add Live Parameter To Search Resource"

sfount:
* partial index on `live` doesn't support falsey check 
* combination of `IN` gateway account id and `live` result in postgres not using indexes, for the `COUNT` (full table) this takes a very long time (would need a combined index `transaction(gateway_account_id, live)`) 

Reverts alphagov/pay-ledger#675


kbottla:
Reverted in https://github.com/alphagov/pay-ledger/pull/678 to unblock master build

PR #717, opened 2020-05-04
PP-6421 Payout entity factory

sfount:
* add `PayoutEntityFactory` to parse payouts from event digests
* simple service method to combine the factory and writing to the DB

Note the object mapper used to convert the event digest payload has to register the java time module (`this.objectMapper.registerModule(new JavaTimeModule())`) otherwise it isn't able to serialise `ZonedDateTime`.

---

What this PR does not do:
* setting state based on PayoutState and known salient events (1 PR)
* setting the event count on the payout to ensure latest event is applied (depends on https://github.com/alphagov/pay-ledger/pull/715)
* setting the payout details to contain anything not at the column level (depends on https://github.com/alphagov/pay-ledger/pull/715)
* handling events at the top level based on resource type (1 PR)

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
+++ b/src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
@@ -1,17 +1,36 @@
 package uk.gov.pay.ledger.payout.entity;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+import uk.gov.pay.ledger.event.model.serializer.MicrosecondPrecisionDateTimeDeserializer;
+import uk.gov.pay.ledger.event.model.serializer.MicrosecondPrecisionDateTimeSerializer;
+
 import java.time.ZonedDateTime;
 
+@JsonIgnoreProperties(ignoreUnknown = true)
+@JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
 public class PayoutEntity {
 
-    private final Long id;
-    private final String gatewayPayoutId;
-    private final Long amount;
-    private final ZonedDateTime createdDate;
-    private final ZonedDateTime paidOutDate;
-    private final String statementDescriptor;
-    private final String status;
-    private final String type;
+    @JsonIgnore
+    private Long id;
+    private String gatewayPayoutId;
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
```diff
@@ -1,17 +1,36 @@
...
+    @JsonIgnore
+    private Long id;
+    private String gatewayPayoutId;
+    private Long amount;
+    @JsonIgnore
+    private ZonedDateTime createdDate;
+    @JsonSerialize(using = MicrosecondPrecisionDateTimeSerializer.class)
+    @JsonDeserialize(using = MicrosecondPrecisionDateTimeDeserializer.class)
+    private ZonedDateTime paidOutDate;
+    private String statementDescriptor;
```
kbottla, line 27: Think this field can go into payout_details json blob ?

src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
```diff
@@ -1,17 +1,36 @@
...
+
+    private String status;
+    private String type;
```
kbottla, line 30: Wonder what `type` is for payout

src/main/java/uk/gov/pay/ledger/payout/service/PayoutService.java
```diff
@@ -0,0 +1,24 @@
...
+    private final PayoutDao payoutDao;
+    private final PayoutEntityFactory payoutEntityFactory;
+
+    @Inject
+    public PayoutService(PayoutDao payoutDao, PayoutEntityFactory payoutEntityFactory) {
+        this.payoutDao = payoutDao;
+        this.payoutEntityFactory = payoutEntityFactory;
+    }
+
+    public void upsertTransactionFor(EventDigest eventDigest) {
```
kbottla, line 20:
```suggestion
    public void upsertPayoutFor(EventDigest eventDigest) {
```

src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
sfount: type is the type of payout, the definition was based on the Stripe payout initially, if it's a payout to a bank account, a credit card etc. Again this PR doesn't add this it just makes it non-final

src/main/java/uk/gov/pay/ledger/payout/service/PayoutService.java
sfount: 🤦

src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
sfount: payout details doesn't yet exist in this PR, I think because the payout will rarely if ever change in size it's okay to have something like this at the top -- we can definitely move it once payout_details is in if that's proposed

PR #723, opened 2020-05-05
PP-6421 Add EventDigestHandler

kbottla:
## WHAT
- Added EventDigestHandler which looks up resource type and processes digest accordingly (i.e, by payout/refund or payout or unsupported)

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
+++ b/src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
@@ -24,20 +24,17 @@ public class EventMessageHandler {
 
     private final EventQueue eventQueue;
     private final EventService eventService;
-    private final TransactionService transactionService;
-    private final TransactionMetadataService transactionMetadataService;
+    private final EventDigestHandler eventDigestHandler;
     private final MetricRegistry metricRegistry;
 
     @Inject
     public EventMessageHandler(EventQueue eventQueue,
                                EventService eventService,
-                               TransactionService transactionService,
-                               TransactionMetadataService transactionMetadataService,
+                               EventDigestHandler eventDigestHandler,
                                MetricRegistry metricRegistry) {
         this.eventQueue = eventQueue;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
```diff
@@ -65,9 +62,7 @@ private void processSingleMessage(EventMessage message) throws QueueException {
         final long ingestLag = event.getEventDate().until(ZonedDateTime.now(), ChronoUnit.MICROS);
 
         if(response.isSuccessful()) {
-            EventDigest eventDigest = eventService.getEventDigestForResource(event.getResourceExternalId());
-            transactionService.upsertTransactionFor(eventDigest);
-            transactionMetadataService.upsertMetadataFor(event);
+            eventDigestHandler.processEvent(event);
```
sfount, line 65: Nice 👍 clean abstraction

src/test/java/uk/gov/pay/ledger/queue/EventDigestHandlerTest.java
```diff
@@ -0,0 +1,116 @@
...
+        eventDigestHandler.processEvent(event);
+
+        verify(transactionService, never()).upsertTransactionFor(any());
+        verify(transactionMetadataService, never()).upsertMetadataFor(any());
+        verify(payoutService, never()).upsertPayoutFor(any());
+
+        verify(mockAppender, times(1)).doAppend(loggingEventArgumentCaptor.capture());
+        List<LoggingEvent> loggingEvents = loggingEventArgumentCaptor.getAllValues();
+        assertThat(loggingEvents.get(0).getFormattedMessage(),
+                is(format("Event digest processing for resource type [%s] is not supported. " +
```
sfount, line 112: Potentially don't need to assert on the exact string logged as that may make it brittle in the future? Maybe a substring match like `resource type is not supported` or something would do. Nothing that needs to change now but can just see this coming up

PR #724, opened 2020-05-05
PP-6421 Payout state with salient events

sfount:
Uniform way of managing external payout state based on salient event
contract (similar to transaction state).

* Add simple lookup in factory to upsert payouts with state based on
their salient events
* Read PayoutState correctly from the database in the payout mapper

---

Mismatch between payout `status` in the database table and the Ledger standard `state` to be addressed in PP-6465.


src/main/java/uk/gov/pay/ledger/payout/state/PayoutState.java
```diff
@@ -0,0 +1,57 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum PayoutState {
+    UNDEFINED("undefined", false),
+    IN_TRANSIT("intransit", false),
+    PAID_OUT("paidout", true),
+    FAILED("failed", true);
+
+    private final String status;
+    private final boolean finished;
```
kbottla, line 22: is there any significance of `finished` for payouts ?

sfount: yeah there are still terminal (completed) and in progress states there just happens to only be one in progress state at the moment! 

Having this in from the beginning will still allow the code downstream to say `if in progress` rather than matching to a specific state.

If we think this we can simplify this further later we can just serialise the JSON to the "status" and remove this param.

kbottla: Makes sense. It's is possible that payout is created but won't be finished for a few hours. Looks useful to have especially if we this available via API

sfount:
This passed on Jenkins -- looks like concourse failed for unrelated reasons.

PR #725, opened 2020-05-05
PP-6467 Transaction CSV resource is case insensitive

sfount:
All metadata keys are treated as lowercase, we validate against this
when creating payments but then show case sensitive columns in the CSV
download. `lower` all of the keys when generating and populating
headers.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
@@ -121,7 +121,7 @@ public class CsvTransactionFactory {
 
             externalMetadata.ifPresent(metadata ->
                     metadata.forEach((key, value) ->
-                            result.put(String.format("%s (metadata)", key), value)));
+                            result.put(String.format("%s (metadata)", key.toLowerCase()), value)));
         } catch (IOException e) {
             LOGGER.error("Error during the parsing transaction entity data [{}] [errorMessage={}]",
                     transactionEntity.getExternalId(), e.getMessage());
@@ -194,7 +194,11 @@ public class CsvTransactionFactory {
         }
 
         if (metadataKeys != null) {
-            metadataKeys.stream().sorted()
+            metadataKeys
+                    .stream()
+                    .map(String::toLowerCase)
+                    .distinct()
+                    .sorted()
                     .forEach(key -> {
                         String header = String.format("%s (metadata)", key);
                         headers.put(header, header);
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
```diff
@@ -116,6 +116,22 @@ public void toMapShouldIncludeExternalMetadataFields() {
...
+                        .put("external_metadata",
+                                ImmutableMap.builder()
+                                        .put("some-key", "value-1").put("SOME-KEY", "value-2").put("Some-key", "value-1").build())
+                        .build()
+                )).toEntity();
+
+        Map<String, Object> csvDataMap = csvTransactionFactory.toMap(transactionEntity);
+
+        assertThat(csvDataMap.get("some-key (metadata)"), is("value-1"));
+        assertThat(csvDataMap.get("SOME-KEY (metadata)"), is(nullValue()));
```
kbottla, line 132: I thought we should only have one header for `some-key` and not have `SOME-KEY`. Looks like this assertion is implying we will have a header with null value. May be change  the assertion to check `some-key` is unique in key set.

src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
```diff
@@ -194,7 +194,11 @@ public CsvTransactionFactory(ObjectMapper objectMapper) {
         }
 
         if (metadataKeys != null) {
-            metadataKeys.stream().sorted()
+            metadataKeys
+                    .stream()
+                    .map(String::toLowerCase)
```
kbottla, line 199: I think this will change all metadata headers to always be in lowercase. Could impact services if their processing is case sensitive. 

Suggested change would be to add to headers like below, so in case on duplicate keys last one will take preference. Otherwise, columns stay unchanged.

`headers.put(header.toLowerCase(), header);`

src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
sfount: Agreed 👍 good suggestion will do that

src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
sfount: I see why this would be beneficial for existing integrations, the problem for the future is that it will be unpredictable -- it would be difficult to make any assertions about what to expect headers to be called if there is a conflict

kbottla: I see. This could be sensitive change in that case. A way could be, if there is a duplicate, it will always be lowered. If not it will be presented as is

sfount:
This draft is associated with a ticket that hasn't yet been prioritised or played by a team with the police cat process. Hopefully it can be used as a pointer for how this might be done but the ticket can assess any remaining questions and get the code through.

PR #726, opened 2020-05-05
PP-6421 Set event_count & payout_details on PayoutEntity

kbottla:
## WHAT
- Sets event_count and payout_details when processing event digest for payouts
- Also updated DAO so we only update payouts when there a new event for payout

with @sandorarpa

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/payout/model/PayoutEntityFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/payout/model/PayoutEntityFactory.java
@@ -1,18 +1,24 @@
 package uk.gov.pay.ledger.payout.model;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/payout/model/PayoutEntityFactory.java
```diff
@@ -1,18 +1,24 @@
...
+import java.util.Map;
+
 public class PayoutEntityFactory {
 
     private final ObjectMapper objectMapper;
+    private static final Logger LOGGER = LoggerFactory.getLogger(PayoutEntityFactory.class);
 
-    public PayoutEntityFactory(){
-        this.objectMapper = new ObjectMapper();
-        this.objectMapper.registerModule(new JavaTimeModule());
```
sfount, line 15: I think this is needed otherwise the payout can't serialise `ZonedDateTime` 🤔

kbottla: Tests are fine 🤷‍♂️

PR #729, opened 2020-05-06
PP-6465 refactor payout entity

SandorArpa:
## WHAT
- drop status (renamed to state in previous PR) column from payout
- drop statement_descriptor and type from payout (not top level data)
- refactor PayoutEntity to use state column instead of status


src/main/java/uk/gov/pay/ledger/payout/state/PayoutState.java
```diff
@@ -18,18 +18,18 @@
     PAID_OUT("paidout", true),
     FAILED("failed", true);
 
-    private final String status;
+    private final String state;
```
sfount, line 21: I believe this should still be `status`, the current public API models looks like: 
```
{
  "state": {
    "status": "the_status"
  }
}
```

PR #739, opened 2020-05-15
PP-6469 Pact test for PayoutCreated event

kbottla:
## WHAT
- Added consumer pact test for PayoutCreated event received from connector (provider)
- Fixes ZonedDateTime format when generating pact as json body, as toString() can give varying output based on input. That is, for date time `2020-05-13T18:45:00.000000Z`, toString() returns `2020-05-13T18:45` instead of full date time as string which breaks https://github.com/alphagov/pay-ledger/blob/f76e13cdf4f369e298f71a80f148355d5d4d6bf5/src/main/java/uk/gov/pay/ledger/event/model/serializer/MicrosecondPrecisionDateTimeSerializer.java#L35

### DIFF
--- a/src/test/java/uk/gov/pay/ledger/util/fixture/QueueEventFixtureUtil.java
+++ b/src/test/java/uk/gov/pay/ledger/util/fixture/QueueEventFixtureUtil.java
@@ -10,6 +10,7 @@ import uk.gov.pay.ledger.event.model.ResourceType;
 import uk.gov.pay.ledger.rule.SqsTestDocker;
 
 import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
 
 public class QueueEventFixtureUtil {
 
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/pay/ledger/util/fixture/QueueEventFixtureUtil.java
```diff
@@ -39,8 +40,10 @@ public static PactDslJsonBody getAsPact(String eventType, ZonedDateTime eventDat
                                      String parentResourceExternalId, ResourceType resourceType, String eventData) {
         PactDslJsonBody eventDetails = new PactDslJsonBody();
 
+        DateTimeFormatter formatToMicroseconds = DateTimeFormatter.ofPattern("uuuu-MM-dd'T'HH:mm:ss.SSSSSSX");
```
sfount, line 43: Is there a reason this couldn't use the static https://github.com/alphagov/pay-ledger/blob/f76e13cdf4f369e298f71a80f148355d5d4d6bf5/src/main/java/uk/gov/pay/ledger/event/model/serializer/MicrosecondPrecisionDateTimeSerializer.java linked in the description?

PR #742, opened 2020-05-18
PP-6502 add gateway_account_id to payouts table

SandorArpa:
## WHAT
- add new column gateway_account_id VARCHAR(255) to payouts table

### DIFF
--- a/src/main/resources/migrations/00051_add_gateway_account_id_to_payout.sql
+++ b/src/main/resources/migrations/00051_add_gateway_account_id_to_payout.sql
@@ -1,0 +1,7 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_gateway_account_id_to_payout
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00051_add_gateway_account_id_to_payout.sql
```diff
@@ -0,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_gateway_account_id_to_payout
+ALTER TABLE payouts ADD COLUMN gateway_account_id VARCHAR(255);
```
sfount, line 4: As this table isn't being used in any environments yet it should be fine to add an index changeset in this migration too. We'll be filtering on `gateway_account_id` so a default index on just this column should be fine.

SandorArpa: 👍

PR #745, opened 2020-05-18
PP-6025 add gateway account id to PayoutEntity

SandorArpa:
## WHAT
- add gateway account id to PayoutEntity
- update tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
+++ b/src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
@@ -28,6 +28,7 @@ public class PayoutEntity {
     private PayoutState state;
     private Integer eventCount;
     private String payoutDetails;
+    private String gatewayAccountId;
 
     public PayoutEntity() {
     }
@@ -41,6 +42,7 @@ public class PayoutEntity {
         this.state = builder.state;
         this.eventCount = builder.eventCount;
         this.payoutDetails = builder.payoutDetails;
+        this.gatewayAccountId = builder.gatewayAccountId;
     }
 
     public Long getId() {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
```diff
@@ -21,13 +21,14 @@
     private String gatewayPayoutId;
     private Long amount;
     @JsonIgnore
-    private ZonedDateTime createdDate;
     @JsonSerialize(using = MicrosecondPrecisionDateTimeSerializer.class)
+    private ZonedDateTime createdDate;
```
sfount, line 25: This looks like it moved created date under the serializer rather than being under `JsonIgnore`, this would change the behaviour of the class

kbottla, line 25: Any reason, `created_date` has now `JsonSerialize` annotation ?

src/main/resources/migrations/00051_add_gateway_account_id_to_payout.sql
```diff
@@ -8,4 +8,4 @@ ALTER TABLE payout ADD COLUMN gateway_account_id VARCHAR(255);
 --changeset uk.gov.pay:add_index_on_gateway_account_id_to_payout_table
 CREATE INDEX payout_gateway_account_id_idx ON payout(gateway_account_id);
 
---rollback drop index payout_gateway_account_id_idx;
\ No newline at end of file
+--rollback drop index payout_gateway_account_id_idx;
```
kbottla, line 11: Not sure if liquibase will break as it calculates checksum for content and does it ignore if it sees `00051_add_gateway_account_id_to_payout.sql` as already migrated ?

src/main/java/uk/gov/pay/ledger/payout/entity/PayoutEntity.java
sfount: Oh no, the same field is being ignored, just looks like it's got a serialise rule now too which I think applies to the field below

PR #752, opened 2020-05-19
PP-6472 Persist and return gateway_payout_id on transaction

stephencdaly:
Add gateway_payout_id column to TransactionEntity and persist it as a
top-level field.

Return gateway_payout_id as field on transactions returned by the get
transactions API endpoints.


src/main/java/uk/gov/pay/ledger/transaction/model/Payment.java
```diff
@@ -85,6 +85,7 @@ public Payment(Long id, String gatewayAccountId, Long amount,
         this.live = live;
         this.source = source;
         this.walletType = walletType;
+        this.gatewayTransactionId = gatewayTransactionId;
```
sfount, line 88: I think this was intending to set the newly added payout id
```suggestion
        this.gatewayPayoutId = gatewayPayoutId;
```

stephencdaly: ah dunno why i did this, it's set in the base class constructor so this line just needs to go

stephencdaly: fixed

sfount: Yeah much simpler 👍 looks good

PR #769, opened 2020-05-26
PP-6585 Calculate parent_external_id from events

kbottla:
## WHAT 

- Derived parent_external_id if any one event in the given list has parent_external_id set
  This allows to send events without parent_external_id as in the case of refunds where payout
  notifications are received from Stripe and connector can send refund related events without looking further into refunds table.
- This also protects against processing new events in future without parent_external_id

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
+++ b/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
@@ -9,6 +9,8 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
+import static org.apache.commons.lang3.StringUtils.isNotEmpty;
+
 public class EventDigest {
     private static final ObjectMapper objectMapper = new ObjectMapper();
     private final ZonedDateTime mostRecentEventTimestamp;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
```diff
@@ -58,18 +60,28 @@ public static EventDigest fromEventList(List<Event> events) {
...
                 eventPayload,
                 earliestDate
         );
     }
 
+    private static String deriveParentResourceExternalId(List<Event> events) {
+        return events.stream()
+                .filter(event -> isNotEmpty(event.getParentResourceExternalId()))
+                .map(Event::getParentResourceExternalId)
+                .findFirst()
```
sfount, line 81: I think Java streams have a `.distinct()` method that would work on strings like this.

We might want to do some warning logging if there are more than one? We could still use the result of `findFirst()` but just make some noise.

kbottla: To log warning it is forcing to use stream filtering twice and method could look complex. will think it through and raise a separate PR as needed.

PR #780, opened 2020-05-28
BAU - Use dockerhub for ryuk image

kbottla:
## WHAT 
ryuk container used by testcontainers library is responsible for starting and cleaning up postgres/sqs containers used during the tests.

By default images is pulled from quay.io.

Added configuration to use dockerhub instead as for all other images. Image is still an official test containers ryuk image https://hub.docker.com/r/testcontainersofficial/ryuk

Custom images can be configured in multiple ways as per https://www.testcontainers.org/features/configuration/


src/test/resources/testcontainers.properties
```diff
@@ -0,0 +1,2 @@
+# Digest of image testcontainersofficial/ryuk@sha256 for linux/amd64 from tags list on https://hub.docker.com/r/testcontainersofficial/ryuk
```
alexbishop1, line 1: Presumably this should identify the version e.g. “Digest of image testcontainersofficial/ryuk:latest …” or “Digest of image testcontainersofficial/ryuk:0.3.0 …”

kbottla: Updated to `:latest`

PR #820, opened 2020-07-01
PP-6708 Upgrade to JUnit5

kbottla:
## WHAT
- Upgrades unit tests to Junit5.

## HOW
- Units tests should pass on CI

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -173,6 +173,12 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit5.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.testcontainers</groupId>
             <artifactId>postgresql</artifactId>
             <version>${testcontainers.version}</version>
@@ -197,7 +203,7 @@
         </dependency>
         <dependency>
             <groupId>org.mockito</groupId>
-            <artifactId>mockito-core</artifactId>
+            <artifactId>mockito-junit-jupiter</artifactId>
             <version>3.3.3</version>
             <scope>test</scope>
         </dependency>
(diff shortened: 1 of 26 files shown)


pom.xml
```diff
@@ -172,6 +172,12 @@
             <version>1.1.1</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <version>${junit5.version}</version>
```
alexbishop1, line 178: Where is this being defined?

src/test/java/uk/gov/pay/ledger/event/service/EventServiceTest.java
```diff
@@ -55,7 +56,7 @@ public void setUp() {
                 .withResourceExternalId(resourceExternalId)
                 .withEventDate(ZonedDateTime.now().minusHours(2L))
                 .toEntity();
-        when(mockEventDao.getEventsByResourceExternalId(resourceExternalId)).thenReturn(List.of(event1, event2));
+        lenient().when(mockEventDao.getEventsByResourceExternalId(resourceExternalId)).thenReturn(List.of(event1, event2));
```
alexbishop1, line 59: Why is this needed now?

pom.xml
kbottla: It is using the one defined in parent POM (dropwizard). Updated in this PR https://github.com/alphagov/pay-ledger/pull/822/files#diff-600376dffeb79835ede4a0b285078036R18

src/test/java/uk/gov/pay/ledger/event/service/EventServiceTest.java
kbottla: Mockito for Junit5 complains about unused stubbing and tests fails, as this stub is only used for few tests. It might be that previous version doesn't require strict stubs or has different rules for it. Going to look into more detail

PR #827, opened 2020-07-06
WIP - Project payment fields onto refund

kbottla:
---- DO NOT MERGE ----
- This PR works on the assumption that `reference` field is only available on payment. At present, it is available on both refunds and payments records. Aim is to move `reference` on refunds payload from connector to `gateway_transaction_id` (provided reference and gateway_transaction_id represent same info in connector refunds table). 

## WHAT 

- Move the responsibility for getting the eventdigest into the process method
- Projects payments fields (required to search for refunds or report (CSV) in refunds) onto refund transaction.
- This change will project payment fields directly on to transaction table (i.e, cardholder_name, email and so on) and additional fields on to transaction_details json blob.


kbottla:
This is done in 
https://github.com/alphagov/pay-ledger/pull/839
https://github.com/alphagov/pay-ledger/pull/840

PR #839, opened 2020-07-17
PP-6657 Split out event digest processors for domain object projection

sfount:
* Move event digest collection up to the processor level
* Model anything that can be projected as an event processor, this
should collect an event digest and project a domain object

with @kbottla
with @heathd

PR #843, opened 2020-07-20
PP-6657 Add shared refund payment details to refund model

sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/839 and https://github.com/alphagov/pay-ledger/pull/840

This will fail until consumer pacts are updated. 

Update refund and transaction view to set shared payment details. This
is set by building the payment pojo in the transaction factory.

Verify this can be parsed through to JSON in a resource response, verify
that it's built correctly given values from the db.


src/main/java/uk/gov/pay/ledger/transaction/search/model/TransactionView.java
```diff
@@ -331,6 +337,7 @@ public String getGatewayPayoutId() {
         private Source source;
         private String walletType;
         private String gatewayPayoutId;
+        private Payment sharedPaymentDetails;
```
kbottla, line 340: Can we TransactionView here similar to `parentTransaction` or a new one ? TransactionView is solely to serialising data and Payment/Refund domain objects are used as a layer in between dao and resource. Using Payment along with TransactionView could mix use cases

PR #844, opened 2020-07-20
PP-6657 Project payment details to nested blob

kbottla:
## WHAT
- Project payment specific information as a nested object `payment_details` on `transaction_details` json blob
- Sets top level column directly instead of using objectMapper to set fields on refund transaction entity

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/entity/TransactionEntity.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/entity/TransactionEntity.java
@@ -8,6 +8,7 @@ import uk.gov.pay.commons.model.Source;
 import uk.gov.pay.ledger.transaction.state.TransactionState;
 
 import java.time.ZonedDateTime;
+import java.util.Map;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
 @JsonNaming(PropertyNamingStrategy.SnakeCaseStrategy.class)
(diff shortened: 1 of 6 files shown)


src/main/java/uk/gov/pay/ledger/transaction/entity/TransactionEntity.java
```diff
@@ -228,6 +229,16 @@ public String getGatewayPayoutId() {
         return gatewayPayoutId;
     }
 
+    public void setFieldsFromOriginalPayment(TransactionEntity paymentTransaction) {
```
sfount, line 232: Reading this, it's a bit hard to understand the semantic difference between `getFieldsFromOriginalPayment` and `setFieldsFromOriginalPayment`. 

`getFieldsFromOriginalPayment` is used to then update the refund event digest payload and `setFieldsFromOriginalPayment` is used separately to update the entity. 

Having the logic in both places means to follow this I have to understand the current method (combining event payloads for both resources) and the proposed method (setting entities).

If this could just read:
```java
mayBePaymentEventDigest.ifPresent(paymentEventDigest -> {
            TransactionEntity paymentTransactionEntity = transactionEntityFactory.create(paymentEventDigest);
            refundTransactionEntity.setFieldsFromOriginalPayment(paymentTransactionEntity);
        });
```

It would seem as clear as the original method, the "set" method would then be responsible for setting the `card_brand_label`, `expiry_date` and `wallet_type`. 

If that's too much work then potentially some more specific naming could make it clearer.

kbottla: Agree naming could be better. Updated to make the names clearer. I was aiming to have simple setters on entity or else it could bloat with responsibilities. If it is something we can do it in more simple way happy to do it in a separate PR.

sfount: Naming changes look good, I think we're okay on "responsibilities" as it's one responsibility is to set fields based on a parent transaction, the only thing I can see right now is that it only does a few of the fields and not the ones that go inside transaction details (which it isn't immediately clear why).

 I agree it seems like something that would be a separate PR.

PR #871, opened 2020-08-04
PP-6768 Add new query param `limit_total`

kbottla:
## WHAT 
- Added new query param `limit_total`. When param is set, counting total for search results is limited to `limit_tota_size` value or defaults.   When `total` is less than `limit_tota_size`, then `total` is accurate, otherwise it should be interpreted as ' > x rows' (where x is `limit_tota_size` set of default)
- Different set of pagination links are generated when `limit_total` is set. last_page is not available as accurate `total` is unknown.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
@@ -10,6 +10,10 @@ public abstract class SearchParams {
 
     public abstract Long getDisplaySize();
 
+    public boolean limitTotal() {
+        return false;
+    }
+
     protected boolean isSet(CommaDelimitedSetParameter commaDelimitedSetParameter) {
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
```diff
@@ -10,6 +10,10 @@
 
     public abstract Long getDisplaySize();
 
+    public boolean limitTotal() {
+        return false;
```
SandorArpa, line 14: Is this an oversight or on purpose?

kbottla: It is required and used here https://github.com/alphagov/pay-ledger/pull/871/files#diff-fe80f02b1b708a029ad3b2a5f520a889R52

PR #878, opened 2020-08-10
PP-6850 Make public method for refund reprojection

stephencdaly:
Make public method in RefundEventProcessor for reprojecting a refund with payment information. This will be called by the
PaymentEventProcessor when a payment that is updated has existing refunds that need to be reprojected with the new payment details.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/queue/eventprocessor/RefundEventProcessor.java
+++ b/src/main/java/uk/gov/pay/ledger/queue/eventprocessor/RefundEventProcessor.java
@@ -33,6 +33,21 @@ public class RefundEventProcessor extends EventProcessor {
         EventDigest refundEventDigest = eventService.getEventDigestForResource(event);
         Optional<EventDigest> mayBePaymentEventDigest = Optional.empty();
 
+        if (isNotBlank(refundEventDigest.getParentResourceExternalId())) {
+            mayBePaymentEventDigest = getPaymentEventDigest(refundEventDigest.getParentResourceExternalId());
+        }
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -152,8 +153,38 @@ public void upsertTransactionFor(EventDigest eventDigest) {
         transactionDao.upsert(transaction);
     }
 
-    public void upsertTransaction(TransactionEntity transaction) {
-        transactionDao.upsert(transaction);
+    public void upsertRefundTransactionWithPaymentInfo(EventDigest refundEventDigest, EventDigest paymentEventDigest) {
```
kbottla, line 156: Previously EventDigestHandler had multiple responsibilities (processing payment, refunds payout events) and has been split into individual processors.
This is moving much of the responsibility of RefundEventProcessor to TransactionService which could be moving us back to previous pattern ? Can we call RefundEventProcessor from PaymentEventProcessor and project refunds ?

Happy to discuss further if this doesn't makes sense

stephencdaly: Not sure this would entirely make sense to do, as we wouldn't be processing a refund event in this case.

Here's a draft PR of how this will be used https://github.com/alphagov/pay-ledger/pull/879

Maybe this logic could live somewhere else that isn't TransactionService, if it's a problem of TransactionService getting too big?

PR #912, opened 2020-09-04
PP-7018 Remove limit for CSV transactions

kbottla:
## WHAT 
- Removed limit on CSV transactions so we can enable and limit total count to 5K for selfservice searches.
  For services with lots of transactions, users will need to apply filters (in selfservice) before downloading transactions.

- Also added .secrets.baseline file which is needed for gds-pre-commit (https://github.com/alphagov/gds-pre-commit) tool

### DIFF
--- a/.secrets.baseline
+++ b/.secrets.baseline
@@ -1,0 +1,75 @@
+{
+  "exclude": {
+    "files": null,
+    "lines": null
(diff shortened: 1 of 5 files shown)


SandorArpa (approved):
LOKTM

PR #917, opened 2020-09-07
PP-7020: Fix env-map db user parameter

rjbaker:
Correctly maps database username from secret service to a local environment variable.


rorymalcolm (approved):
🥍

PR #919, opened 2020-09-08
PP-6669 add payout date to search transaction with cursor

SandorArpa:
## WHAT
- join payout table on gateway payout id to transactions when cursor searching transactions
- add a relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
@@ -18,14 +18,14 @@ import java.util.stream.Collectors;
 import static org.apache.commons.lang3.StringUtils.isBlank;
 
 public class TransactionDao {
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -70,7 +70,10 @@
             "ORDER BY t.created_date DESC OFFSET :offset LIMIT :limit";
 
     private static final String SEARCH_TRANSACTIONS_CURSOR =
-            "SELECT * FROM transaction t " +
+            "SELECT t.*, po.paid_out_date AS paid_out_date FROM transaction t " +
+            "LEFT OUTER JOIN payout po on " +
+            "t.gateway_payout_id = po.gateway_payout_id " +
+            ":payoutAccountId " +
```
kbottla, line 76: Can we name \:payoutAccountId to something different (\:oayoutSearchFields or ..) to avoid confusing it as bind variable for field rather than join condition. For a moment I got confused and read it as field (as we do this for all other fields like \:gatewayAccountId, \:externalId) rather than condition

kbottla (approved):

PR #925, opened 2020-09-11
PP-6669 add settlement summary to refund

SandorArpa:
## WHAT
- add settlement summary to refund transaction type
- refactor SettlementSummary to have one common field between the different types of transactions
- add PaymentSettlementSummary that extends SettlementSummary
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/TransactionFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/TransactionFactory.java
@@ -7,8 +7,9 @@ import com.google.inject.Inject;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
-import uk.gov.pay.ledger.transaction.search.model.RefundSummary;
 import uk.gov.pay.ledger.transaction.search.model.SettlementSummary;
+import uk.gov.pay.ledger.transaction.search.model.RefundSummary;
+import uk.gov.pay.ledger.transaction.search.model.PaymentSettlementSummary;
 
 import java.io.IOException;
 import java.util.Map;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/TransactionFactory.java
```diff
@@ -125,6 +125,12 @@ private Transaction createRefund(TransactionEntity entity) {
                     .withWalletType(safeGetAsString(refundPaymentDetails, "wallet"))
                     .build();
 
+            SettlementSummary settlementSummary = new SettlementSummary(
+                    safeGetAsDate(transactionDetails, "capture_submitted_date"),
+                    safeGetAsDate(transactionDetails, "captured_date"),
+                    entity.getPayoutEntity().map(payoutEntity -> payoutEntity.getPaidOutDate()).orElse(null)
```
kbottla, line 131: Can we build with only paid out date as we don't need to look for `capture_submitted_date` and `captured_date` for refund ?

src/main/java/uk/gov/pay/ledger/transaction/model/TransactionFactory.java
```diff
@@ -138,6 +144,7 @@ private Transaction createRefund(TransactionEntity entity) {
                     .withParentExternalId(entity.getParentExternalId())
                     .withGatewayPayoutId(entity.getGatewayPayoutId())
                     .withPaymentDetails(paymentDetails)
+                    .withSettlementSummary(settlementSummary)
```
kbottla, line 147: What will be the behaviour when paidout_date is not available ? do we get `settlement_summary` without any fields ?

src/main/java/uk/gov/pay/ledger/transaction/search/model/SettlementSummary.java
```diff
@@ -1,74 +1,4 @@
...
-    }
-
-    @Override
-    public String toString() {
-        return "SettlementSummary{" +
-                ", settlementSubmittedTime=" + settlementSubmittedTime +
-                ", capturedDate=" + capturedDate +
-                '}';
-    }
+public abstract class SettlementSummary {
```
kbottla, line 3: Empty abstract class could cause confusion as there is no functionality/behaviour inherited here except that this is used for Jackson. We could potentially have settled_date in SettlementSummary and PaymentSettlementSummary could inherit this. For refunds, we can use SettlementSummary directly.

kbottla (approved):
LGTM!

PR #946, opened 2020-09-25
BAU fix query bug for null gateway account ids

SandorArpa:
## WHAT
- fix the query for single transaction when no gateway account id is present


sfount (approved):
It would be good to have a test that falls over if this set wrong but looks good to me otherwise 👍

SandorArpa:
closing as no corresponding JIRA ticket was created

PR #947, opened 2020-09-26
PP-7223 fix query bug for priviliged accounts

SandorArpa:
## WHAT
- fix query bug when joining payout table to transactions with no gateway account id is present


kbottla (approved):
LGTM!

PR #954, opened 2020-09-30
PP-7241 Add `paas` environment prefix

danworth:
The `ENVIRONMENT` variable is used to distinguish our various
environments in Sentry (and is added to our structured logging under the
environment key).

This change will only impact apps deployed on PaaS and should allow us
to separate issues more obviously in Sentry.

## WHAT ##
Strictly speaking existing infra and paas apps have different `environment` tags in Sentry because paas environments do not have the `n` suffix, but this makes it much clearer.

### DIFF
--- a/manifest.yml
+++ b/manifest.yml
@@ -36,7 +36,7 @@ applications:
 
       ADMIN_PORT: '10701'
       DISABLE_INTERNAL_HTTPS: ((disable_internal_https))
-      ENVIRONMENT: ((space))
+      ENVIRONMENT: paas-((space))
       JAVA_OPTS: -Xms512m -Xmx1G
       JBP_CONFIG_JAVA_MAIN: '{ arguments: "server /home/vcap/app/config/config.yaml" }'
       JBP_CONFIG_OPEN_JDK_JRE: '{ jre: { version: 11.+ } }'


katstevens:
Is `ENVIRONMENT` just used by Sentry at the moment then?

danworth:
It's also used by our logger to add the `environment` key/value. This configuration lives in here if you search for `ENVIRONMENT`.
https://github.com/alphagov/pay-ledger/blob/master/src/main/resources/config/config.yaml#L12

The first use is by the request logger which logs a line for each request as the name suggests
https://github.com/alphagov/pay-ledger/blob/2ba06c39a7a5a99cfe0a5ff47f765044029b82f4/src/main/resources/config/config.yaml#L12

The second use is by the general logger so it can add `environment: <value>` to each log entry
https://github.com/alphagov/pay-ledger/blob/2ba06c39a7a5a99cfe0a5ff47f765044029b82f4/src/main/resources/config/config.yaml#L56

And finally the third use is by Sentry to add an environment tag to allow us to filter/search etc.
https://github.com/alphagov/pay-ledger/blob/2ba06c39a7a5a99cfe0a5ff47f765044029b82f4/src/main/resources/config/config.yaml#L60

At present the logs from our PaaS apps go to a different index in Splunk so they're unlikely to be confused with existing apps as we're seeing in Sentry. However that might change in which case it'll help both differentiate logs in Splunk and issues in Sentry and it does no harm in the mean time.

This will only impact our PaaS apps since our existing infra makes no use of the `manifest.yml`

katstevens:
Makes sense - thanks for the detailed reply! 👍

katstevens (approved):
Looks good to me!

rorymalcolm (approved):
lgtm 😎

PR #965, opened 2020-10-07
BAU Revert Junit Jupiter bump to 5.7

danworth:
This reverts commit 6788e23ab4fe165b667b6d4affbeb0d49823c466.

This version bump meant that running the consumer contract tests no longer
resulted in the pacts being created in `target/pacts` or AFAIK anywhere.
This meant they could not be pushed to the pact broker and deploying to
environments failed since CI was unable to tag a version of Ledger which
the pact broker had no knowledge of.

## WHAT ##
Tested this out locally, see pacts https://pact-broker-test.cloudapps.digital/hal-browser/browser.html#/pacticipants/ledger/versions/d8b1c7d62c7a6e45ba6b761bbf80fb1701cca3f4

I haven't looked at why this Junit Jupiter version bump was stopping the pacts being created.

Now that we know ledger should have a pact we might want to remove the `do not fail if there are no pacts` flag when running ledger consumer contract tests - this meant this step was failing silently and the problem surfacing during the `can-i-deploy` check upon deployment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -182,13 +182,7 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
-            <version>5.7.0</version>
-            <scope>test</scope>
-        </dependency>
-        <dependency>
-            <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-engine</artifactId>
-            <version>${junit5.version}</version>
+            <version>5.6.2</version>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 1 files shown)


danworth:
It has published the pacts as part of the PR build, seen here: https://pay-concourse-pact-broker.cloudapps.digital/hal-browser/browser.html#/pacticipants/ledger/versions/eb7d4f0c9bbbc2e1e1cc03935d195f4fa371e9a0

dj-maisy (approved):
Yep, pom file looks correct, and I can confirm that rolling this back to `5.6.2` allowed the pact results to be published again.

PR #978, opened 2020-10-15
PP-7276 Add user email collected pact test

kbottla:
## WHAT 
- Added pact test for new event sent by connector when email is patched


stephencdaly (approved):
lgtm

PR #982, opened 2020-10-19
PP-6989 Map `cancel by external service failed` salient event to transaction state

kbottla:
## WHAT
- Maps salient event `CANCEL_BY_EXTERNAL_SERVICE_FAILED` to transaction state `CANCELLED`


sfount (dismissed):
This lines up with Connectors external mapping 🤨 . Good catch

sfount (approved):

PR #1069, opened 2021-01-11
PP-7609 - Refactor lambda into method references

nimalank7:
Description:
- Removes lambdas with method references
- With @alexbishop1

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
+++ b/src/main/java/uk/gov/pay/ledger/event/model/EventDigest.java
@@ -51,7 +51,8 @@ public class EventDigest {
                 .orElseThrow(() -> new EmptyEventsException("No events found"));
 
         var latestSalientEventType = events.stream()
-                .map(e -> SalientEventType.from(e.getEventType()))
+                .map(Event::getEventType)
+                .map(SalientEventType::from)
                 .flatMap(Optional::stream)
                 .findFirst()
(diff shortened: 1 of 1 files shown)


src/main/java/uk/gov/pay/ledger/event/model/SalientEventType.java
```diff
@@ -16,6 +16,7 @@
     UNEXPECTED_GATEWAY_ERROR_DURING_AUTHORISATION,
     STATUS_CORRECTED_TO_AUTHORISATION_ERROR_TO_MATCH_GATEWAY_STATUS,
     GATEWAY_REQUIRES_3DS_AUTHORISATION,
+    GATEWAY_3DS_EXEMPTION_RESULT_OBTAINED,
```
sfount, line 19: Modelling this in Ledger is only needed if this event will map to an external state transition for the entity, this allows the public API responses to be guaranteed consistent independent of consumers emitting events.

For events that don't need to transition the external state of their entity, Ledger can happily parse any name/ data combination without it being explicitly defined.

Not sure of what the use case for the proposed event is so just double checking!

nimalank7: Ha good spot! Removed that changed so its all about refactoring the lambdas into method references.

oswaldquek (approved):

PR #1071, opened 2021-01-12
PP-7609 - Pact test for GATEWAY_3DS_EXEMPTION_RESULT_OBTAINED

nimalank7:
Description:
- This PR creates the JSON file which is needed on the consumer side to verify the pact
- With @alexbishop1


alexbishop1 (approved):
I paired with Nim on this and am confident it needs no further review per our policy.

PR #1072, opened 2021-01-12
PP-7609 - Rename externalId so that it is 26 characters

nimalank7:
Description:
- This PR renames the externalId in Gateway3dsExemptionResultObtainedEventQueueContractTest to exemptionResult_externalId so that it is 26 characters. The previous version was 44 characters which caused a PSQLException when the Pact test was run


alexbishop1 (approved):
🤦‍♂️

PR #1092, opened 2021-02-01
PP-7696 Replace JUnit 4 @Ignore annotation with JUnit 5 @Disabled

alexbishop1:
`TransactionDaoIT` uses JUnit 5.

Also remove the access modifiers on the class and method to make the test more obviously JUnit 5.


SandorArpa (approved):
LOKTM

PR #1093, opened 2021-02-01
PP-7696 Add Source.CARD_AGENT_INITIATED_MOTO and unignore test

alexbishop1:
Update to latest Pay Java commons so that we get the updated `Source` enum with the `CARD_AGENT_INITIATED_MOTO` value. The database was migrated to be aware of this in commit 13233115e27dd789596a107443f3d57beb6d45a9.

Re-enable the test that checks the `Source` enum in the code matches the `source` enum in the database because they’re in sync again now.


SandorArpa (approved):
LGTM

PR #1126, opened 2021-03-02
PP-PP-7831 add value column to transaction_metadata table

SandorArpa:
## WHAT
- add a new column `value` to transaction_metadata table
- add btree index to transaction_metadata on lower(value)

### DIFF
--- a/src/main/resources/migrations/00056_index_transaction_metadata_value_column.sql
+++ b/src/main/resources/migrations/00056_index_transaction_metadata_value_column.sql
@@ -1,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_index_transaction_metadata_value_column runInTransaction:false
+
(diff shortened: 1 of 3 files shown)


src/main/resources/migrations/00056_index_transaction_metadata_value_column.sql
```diff
@@ -0,0 +1,6 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_index_transaction_metadata_value_column runInTransaction:false
+
+CREATE INDEX CONCURRENTLY IF NOT EXISTS transaction_metadata_value_column_idx
```
kbottla, line 5: Can we rename index to suggest lower similar to `transaction_lower_reference_idx` ?

```suggestion
CREATE INDEX CONCURRENTLY IF NOT EXISTS transaction_metadata_lower_value_idx
```

kbottla (approved):

PR #1129, opened 2021-03-03
PP-7833 Metadata keys for search transactions by metadata value

kbottla:
## WHAT 
- Changes to find metadata keys for transactions searched by metadata value.
- Currently all metadata keys are returned for matching transactions (based on search criteria). transaction table is joined with transaction_metadata to find all keys.
- When transactions are searched by metadata value, all metadata keys are returned for transactions for which atleast one metadata value matches. For this, first find transactions for metadata value and then find all keys for these transactions.

- Rename `TransactionMetadataKeyDaoIT` to `MetadataKeyDaoIT` as this dao is for operations on metadata_key table only.


SandorArpa (approved):
LGTM

PR #1130, opened 2021-03-03
PP-7832 add upsert method to TransactionMetadataDao

SandorArpa:
## WHAT
- add a new method to upsert values on transaction_metada table
- add relevant test


kbottla (approved):
LGTM!

PR #1134, opened 2021-03-04
PP-7832 replace use of insertIfNotExists with upsert

SandorArpa:
## WHAT
- use upsert in place of insertIfNotExist in TransactionMetadataService
- remove insertIfNotExist method from TransactionMetadataService
- remove insertIfNotExist tests
- update all other tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
@@ -46,11 +46,11 @@ public class TransactionMetadataService {
         if (eventDataNode != null && eventDataNode.has("external_metadata")) {
             transactionDao.findTransactionByExternalId(event.getResourceExternalId())
                     .ifPresent(transactionEntity -> {
-                        Iterator<String> metadataKeys = eventDataNode.get("external_metadata").fieldNames();
-                        metadataKeys.forEachRemaining(metadataKey -> {
-                            metadataKeyDao.insertIfNotExist(metadataKey);
+                        eventDataNode.get("external_metadata").fields().forEachRemaining(metadata -> {
+                            metadataKeyDao.insertIfNotExist(metadata.getKey());
+                            String value = metadata.getValue().textValue();
                             transactionMetadataDao
-                                    .insertIfNotExist(transactionEntity.getId(), metadataKey);
+                                    .upsert(transactionEntity.getId(), metadata.getKey(), value);
                         });
                     });
         }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
```diff
@@ -49,8 +49,9 @@ public void upsertMetadataFor(Event event) {
                         Iterator<String> metadataKeys = eventDataNode.get("external_metadata").fieldNames();
                         metadataKeys.forEachRemaining(metadataKey -> {
                             metadataKeyDao.insertIfNotExist(metadataKey);
+                            String value = eventDataNode.get("external_metadata").get(metadataKey).textValue();
```
stephencdaly, line 52: We need to get the metadata from the EventStream so that we can re-project old transactions to back-fill.
This will work for new transactions though, will using the EventStream be the next step?

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
```diff
@@ -49,8 +49,9 @@ public void upsertMetadataFor(Event event) {
                         Iterator<String> metadataKeys = eventDataNode.get("external_metadata").fieldNames();
```
stephencdaly, line 49: Would it be slightly cleaner to change this to iterate over the `fields()` of the JsonNode so we don't have to get the field by the key again when we iterate?

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
SandorArpa: It's small iterations, next will be reworking the method to accept an EventDigest object and work from that.

stephencdaly: 👍

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
SandorArpa: I have considered this approach but if we iterate over `fields()` then we still need to get the relevant key for the upsert.

stephencdaly: This is what I was thinking:

```
                        eventDataNode.get("external_metadata").fields().forEachRemaining(metadata -> {
                            metadataKeyDao.insertIfNotExist(metadata.getKey());
                            String value = metadata.getValue().textValue();
                            transactionMetadataDao
                                    .upsert(transactionEntity.getId(), metadata.getKey(), value);
                        });
```

stephencdaly (approved):

stephencdaly (approved):

PR #1135, opened 2021-03-04
PP-7832 refactor TransactionMetadataService to use EventDigest for upsert

SandorArpa:
## WHAT

- refactor TransactionMetadataService to use EventDigest instead of Event for upserting transaction_metadata
- add/refactor relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
@@ -1,20 +1,17 @@
 package uk.gov.pay.ledger.transaction.service;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
-import com.fasterxml.jackson.databind.JsonNode;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.inject.Inject;
 import jersey.repackaged.com.google.common.base.Stopwatch;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import uk.gov.pay.ledger.event.model.Event;
+import uk.gov.pay.ledger.event.model.EventDigest;
 import uk.gov.pay.ledger.metadatakey.dao.MetadataKeyDao;
 import uk.gov.pay.ledger.transaction.dao.TransactionDao;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
```diff
@@ -35,25 +32,19 @@ public TransactionMetadataService(MetadataKeyDao metadataKeyDao,
...
-                            metadataKeyDao.insertIfNotExist(metadata.getKey());
-                            String value = metadata.getValue().textValue();
-                            transactionMetadataDao
-                                    .upsert(transactionEntity.getId(), metadata.getKey(), value);
-                        });
-                    });
-        }
+                        if (eventData.containsKey("external_metadata")) {
+                            var metadata = (Map<String, Object>) eventData.get("external_metadata");
+                            metadata.keySet().forEach(key -> {
```
stephencdaly, line 42: Would this be better using `metadata.entrySet()` to match what we were doing before?

src/test/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataServiceTest.java
```diff
@@ -76,9 +84,36 @@ public void shouldNotTryToInsertMetadata() {
...
+        Event previousEvent = anEventFixture().withResourceType(PAYMENT)
+                .withEventType("USER_APPROVED_FOR_CAPTURE")
+                .toEntity();
+        EventDigest paymentEventDigest = EventDigest.fromEventList(List.of(event, previousEvent));
+
+        when(mockTransactionDao.findTransactionByExternalId(externalId)).thenReturn(Optional.of(transaction));
+
+        service.upsertMetadataFor(paymentEventDigest);
+
+        verify(mockTransactionMetadataDao).upsert(any(), anyString(), anyString());
```
stephencdaly, line 117: does it make sense to assert on the actual strings to make extra sure? don't see why we need to use anyString() here.

stephencdaly (commented):
Looks good! Some very minor comments.

stephencdaly (approved):
LGTM!

PR #1140, opened 2021-03-08
PP-7832 add unique constraint on key on metadata_key

SandorArpa:
## WHAT
- add a unique constraint on key column on metadata_key table

### DIFF
--- a/src/main/resources/migrations/00058_make_key_on_metadata_key_unique.sql
+++ b/src/main/resources/migrations/00058_make_key_on_metadata_key_unique.sql
@@ -1,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_unique_constraint_on_key_on_metadata_key
+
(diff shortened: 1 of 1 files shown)


src/main/resources/migrations/00058_make_key_on_metadata_key_unique.sql
```diff
@@ -0,0 +1,5 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_unique_constraint_on_key_on_metadata_key
+
+ALTER TABLE metadata_key ADD CONSTRAINT key_unique_key UNIQUE (key);
```
kbottla, line 5: It is tricky with the naming for this. May be the below `metadata_key_unique_key` or `metadata_key_unique` ?
 ```suggestion
ALTER TABLE metadata_key ADD CONSTRAINT metadata_key_unique_key UNIQUE (key);
```

kbottla (approved):

PR #1147, opened 2021-03-11
PP-7859 Insert transaction_metadata for metadata pact state

kbottla:
## WHAT
- For pact state `a transaction with metadata exists`, insert transaction_metadata & metadata rows as well so this state can be used by pact (search by metadata)


SandorArpa (approved):
LOKTM

PR #1152, opened 2021-03-16
PP-7906 Reproject transaction_metadata

stephencdaly:
Reproject `transaction_metadata` rows from the EventDigest when an event is processed with the field `reproject_domain_object` set to true.

This will not take into account any of the event details on the event being processed, only previous events for the transaction from the database.

When an event is processed without this field, or set to false, continue to only upsert to the `transaction_metadata` when the event being processed has `external_metatada` in its event details.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
@@ -9,12 +9,15 @@ import org.jdbi.v3.core.statement.UnableToExecuteStatementException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import uk.gov.pay.ledger.event.model.Event;
+import uk.gov.pay.ledger.event.model.EventDigest;
 import uk.gov.pay.ledger.metadatakey.dao.MetadataKeyDao;
 import uk.gov.pay.ledger.transaction.dao.TransactionDao;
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
```diff
@@ -62,7 +64,21 @@ public void upsertMetadataFor(Event event) {
...
 
-    public List<String> findMetadataKeysForTransactions(TransactionSearchParams searchParams) {
+    public void reprojectFromEventDigest(EventDigest eventDigest) {
+        var eventData = eventDigest.getEventPayload();
+
+        transactionDao.findTransactionByExternalId(eventDigest.getResourceExternalId())
+                .ifPresent(transactionEntity -> {
+                    @SuppressWarnings("unchecked")
+                    var metadata = (Map<String, Object>) eventData.get("external_metadata");
+                    metadata.forEach((key, value) -> {
```
kbottla, line 74: Can we check for null as external_metadata won't be available all the time and add relevant test ?

kbottla (changes requested):

src/main/java/uk/gov/pay/ledger/transaction/service/TransactionMetadataService.java
stephencdaly: 🤦 Good spot! Updated

kbottla (approved):
LGTM!

PR #1156, opened 2021-03-18
PP-7906 Log event_type for reprojection events

stephencdaly:
If event has `reproject_domain_object` flag set to true, log the
event_type so that there is an audit record for when reprojection is
performed.

Also only log the `reproject_domain_object_event` field on the log line
when it is true to avoid unnecessary information on other log lines.


kbottla (approved):
LGTM!

PR #1200, opened 2021-04-26
PP-8025 Update/Add DAO methods for transaction_summary

kbottla:
## WHAT 
- Split out updating fee from transaction_summary insert/update as fee won't be available on the first successful event (which is used to update total_amount_in_pence).
- as fee is available only on capture_confirmed event, updateFee method will be used when processing CAPTURE_CONFIRMED event or when all info to project fee is available.


SandorArpa (approved):
LOKTM

PR #1201, opened 2021-04-26
PP-8026 - Refactor monthly gateway performance report with transaction summary

nimalank7:
Description:
- Refactors /gateway-performance-report to use the transaction_summary table via TransactionSummaryDao
- Since the transaction_summary table doesn't contain a minimum and maximum amount nor can it be calculated from the table it is omitted in the JSON response
- The transaction_summary table no longer has timezone so where appropriate ZonedDateTime has been replaced with LocalDateTime
- The corresponding Toolbox PR for sending dates instead of timezones will need to be merged as LocalDate.parse(...) will throw an exception if timezones are contained. Currently Toolbox sends
timezone data
- Unnecessary .log() from previous testing has been removed

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDao.java
+++ b/src/main/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDao.java
@@ -2,9 +2,13 @@ package uk.gov.pay.ledger.transactionsummary.dao;
 
 import com.google.inject.Inject;
 import org.jdbi.v3.core.Jdbi;
+import uk.gov.pay.ledger.report.entity.GatewayAccountMonthlyPerformanceReportEntity;
+import uk.gov.pay.ledger.report.mapper.GatewayAccountMonthlyPerformanceReportEntityMapper;
 import uk.gov.pay.ledger.transaction.state.TransactionState;
 
+import java.time.LocalDate;
 import java.time.ZonedDateTime;
+import java.util.List;
 
 public class TransactionSummaryDao {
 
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDaoIT.java
```diff
@@ -145,4 +154,70 @@ public void shouldUpdateFeeForTransactionSummaryCorrectly() {
         assertThat(transactionSummary.get(0).get("total_amount_in_pence"), is(1000L));
         assertThat(transactionSummary.get(0).get("total_fee_in_pence"), is(123L));
     }
+
+    @Test
+    public void verifyMonthlyGatewayPerformanceReportTest() {
+        aTransactionSummaryFixture()
+                .withGatewayAccountId("1")
```
kbottla, line 161: Could use random strings like `String gatewayAccountId1 = "account-" + randomAlphanumeric(10);` so multiple tests can avoid using same IDs

src/test/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDaoIT.java
```diff
@@ -145,4 +154,70 @@ public void shouldUpdateFeeForTransactionSummaryCorrectly() {
...
+        List<GatewayAccountMonthlyPerformanceReportEntity> performanceReport = transactionSummaryDao.monthlyPerformanceReportForGatewayAccounts(startDate, endDate);
+
+        GatewayAccountMonthlyPerformanceReportEntity gatewayAccountOnePerformanceReport = performanceReport.get(0);
+        GatewayAccountMonthlyPerformanceReportEntity gatewayAccountTwoPerformanceReport = performanceReport.get(1);
+
+        assertThat(performanceReport.size(), CoreMatchers.is(2));
+
+        assertThat(gatewayAccountOnePerformanceReport.getGatewayAccountId(), CoreMatchers.is(1L));
+        assertThat(gatewayAccountOnePerformanceReport.getTotalVolume(), CoreMatchers.is(1L));
+        assertThat(gatewayAccountOnePerformanceReport.getTotalAmount(), CoreMatchers.is(closeTo(expectedValue, ZERO)));
```
kbottla, line 211: is there any reason `closeTo` is used to assert rather than exact match ?

src/test/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDaoIT.java
nimalank7: I'm not sure what this gains as the data is truncated anyways. It could be the case that the `randomAlphanumeric` function returns identical gateway accounts.

src/main/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDao.java
```diff
@@ -41,6 +45,20 @@
             " AND ts.live = :live" +
             " AND ts.moto = :moto";
 
+    private static final String MONTHLY_GATEWAY_ACCOUNT_PERFORMANCE_STATISTICS = "SELECT " + "t.gateway_account_id, " +
+            "COALESCE(COUNT(t.no_of_transactions), 0) AS volume, " +
```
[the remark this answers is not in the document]
nimalank7: Good spot.

src/test/java/uk/gov/pay/ledger/transactionsummary/dao/TransactionSummaryDaoIT.java
nimalank7: Ah I copied and pasted these tests from PerformanceReportDao. Yes `closeTo` shouldn't be used with getTotalAmount but it has to be used for getAverageAmount as it returns 1000.000... and the expected value is 1000. Since both are BigDecimal they won't be equal so I had to use `closeTo`

kbottla (approved):

PR #1205, opened 2021-04-27
PP-8025 Project transaction summary from event stream

kbottla:
## WHAT
- Projects transaction summary from event stream when transaction state is finished
	- To be able to project transaction summary, PAYMENT_CREATED event should exists for the transaction as this event carries required information (gateway_account_id, moto, live, amount) along with a terminal event. So projects transaction summary if event is PAYMENT_CREATED event and transaction state is finished or for an event that maps to transaction finished state.
	- In case of multiple events (ex: USER_APPROVED_FOR_CAPTURE, CAPTURE_SUBMITTED, CAPTURED_CONFIRMED) leading to same terminal state, transaction_summary is updated only once to avoid updating summary multiple times for a payment.
- Projects fee for CAPTURE_CONFIRMED event as fee is expected only on this event. Multiple CAPTURE_CONFIRMED events with different event timestamps are ignored in this case.


SandorArpa (approved):
LOKTM

SandorArpa (approved):
LGTM

PR #1206, opened 2021-04-27
PP-8025 Fix flakey test

kbottla:
## WHAT
- Increase wait time so messages from queue are processed by queue message handler before asserting on the results.


SandorArpa (dismissed):
LGTM

SandorArpa (approved):
LEBTM

PR #1210, opened 2021-04-28
PP-8050 Use LocalDate for transaction summary

kbottla:
# WHAT
- Use LocalDate explicitly for operations on transaction summary instead of using ZonedDateTime as granualarity of data (in transaction_summary) is at day level


SandorArpa (approved):
LOKTM

PR #1221, opened 2021-05-05
PP-8050 Project transaction_summary for new events only

kbottla:
## WHAT
- Ledger can receive event multiple times for various reasons (parity checker reemitting events, emitting events for a charge for manual fixes, backfilling charges for any reason). In case ledger receives event more than once we don't want to project transaction summary so only project if ledger sees event for the first time.
- Currently summary is projected only when PAYMENT_CREATED event exists but for telephone payment reported the equivalent is "PAYMENT_NOTIFICATION_CREATED" event.
- Excludes CAPTURE_SUBMITTED & CAPTURE_CONFIRMED events as the prior events USER_APPROVED_FOR_CAPTURE and SERVICE_APPROVED_FOR_CAPTURE lead to same transaction state.
- Project transaction summary based on the current working list of events as getting events from database could result in new events (processed by ledger) not seen when projecting transaction.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transactionsummary/service/TransactionSummaryService.java
+++ b/src/main/java/uk/gov/pay/ledger/transactionsummary/service/TransactionSummaryService.java
@@ -3,8 +3,8 @@ package uk.gov.pay.ledger.transactionsummary.service;
 import com.google.inject.Inject;
 import uk.gov.pay.ledger.event.model.Event;
 import uk.gov.pay.ledger.event.model.SalientEventType;
-import uk.gov.pay.ledger.event.service.EventService;
 import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
+import uk.gov.pay.ledger.transaction.model.TransactionType;
 import uk.gov.pay.ledger.transaction.state.TransactionState;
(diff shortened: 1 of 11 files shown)


SandorArpa (approved):
LOKTM

src/main/java/uk/gov/pay/ledger/transactionsummary/service/TransactionSummaryService.java
```diff
@@ -16,65 +16,69 @@
...
-
-        if (!hasPaymentCreatedEvent) {
+        if (!hasPaymentCreatedOrNotificationEvent(events)) {
             return;
         }
 
-        List<Event> eventsMappingToFinishedState =
-                getEventsMappingToTransactionFinishedStateInDescendingOrder(events);
-        Optional<SalientEventType> mayBeCurrentSalientEventType = from(currentEvent.getEventType());
+        if (!ignoreEventsForTransactionAmount.contains(currentEvent.getEventType())) {
```
stephencdaly, line 52: Is it worth adding a comment about why we ignore these payment types? From the code it isn't obvious so could trip someone up later.

kbottla: 👍🏽, will add in next PR

PR #1222, opened 2021-05-05
PP-8050 Use CAPTURE_SUBMITTED event to project transaction summary

kbottla:
## WHAT
- Current transaction summary (for success state) is projected for USER_APPROVED_FOR_CAPTURE OR SERVICE_APPROVED_FOR_CAPTURE events. But for telephone payments these events are not available.
- Use CAPTURE_SUBMITTED event instead as it is available for all use cases (user approved, service approved, telephone payments)


SandorArpa (approved):
LOKTM

PR #1227, opened 2021-05-10
PP-8050 Refactor Transaction summary service

kbottla:
## WHAT
- Refactored transaction summary service to check for conditions separately to project transaction amount and fee.
- Added `event_type` to event logging. It is not clear currently the event type being processed by EventMessageHandler.


stephencdaly (approved):
Looks good

PR #1230, opened 2021-05-12
PP-8047 add github action

SandorArpa:
## WHAT
- add github action to run unit tests on PR creation and/or push (including merging to main)

with @stephencdaly


stephencdaly (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay

PR #1232, opened 2021-05-12
PP-8047 Add dependabot configuration for Github actions

stephencdaly:
Add dependabot configuration to raise pull requests to update GitHub actions when there are new versions. as per https://docs.github.com/en/code-security/supply-chain-security/keeping-your-actions-up-to-date-with-dependabot

with @SandorArpa


kbottla (approved):
LGTM!

PR #1246, opened 2021-05-21
PP-8139 Ledger legacy aggregate endpoint for payments dashboard

sfount:
The live payments dashboard reports on daily aggregates as payments are
being made. Currently the transaction summary table only project
payments that have reached a terminal state, that means that
payments will only be included in the query after they have either
succeeded, failed or expired.

The live payments dashboard never reports on aggregate data over an
individual day, this should never run into performance problems.

Revert the performance report dao and add a `performance-report-legacy`
route to allow the dashboard to continue retrieving up-to-date aggregate
data based on the transaction table for both terminal and in-flight
transactions.


stephencdaly (approved):
Looks good

PR #1288, opened 2021-06-29
PP-8286 Add "Payment Provider" column to CSV stream

kbottla:
## WHAT
- Added new column "Payment Provider" to the CSV stream. Field will be added as last field but before metadata fields
- Only available for payment transactions, as payment provider is not available for refund transactions


stephencdaly (approved):
LGTM

PR #1298, opened 2021-07-07
PP-8270 Add `credential_external_id` to pact tests

kbottla:
## WHAT 
- Added new field `credential_external_id` expected for PAYMENT_CREATED and PAYMENT_NOTIFICATION_CREATED event pacts


SandorArpa (approved):
LOKTM

PR #1366, opened 2021-08-20
PP-8440 Add service_id and live to event, transaction, payout tables

mrlumbu:
We are re-orienting all payment data around services rather than gateway accounts. To enable this, events coming in to ledger will have a top level attribute of `service_id` and `live`.

This adds necessary columns to store these values in event table. I have been consistent with column naming by using
`live` to indicate whether an event is live or test.

Also adds service_id column to transaction table, and adds index on `service_id`. Added concurrently to avoid long lived table lock.

Also adds service_id and live column to payouts, with same index as transaction table,  plus partial index on live payoputs for our old friend toolbox.

### DIFF
--- a/src/main/resources/migrations/00065_add_service_id_column_to_transaction_table.sql
+++ b/src/main/resources/migrations/00065_add_service_id_column_to_transaction_table.sql
@@ -1,0 +1,11 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_service_id_to_transaction_table
+ALTER TABLE transaction
+    ADD COLUMN service_id VARCHAR(32);
+
+--rollback ALTER TABLE transaction DROP COLUMN service_id;
+
+--changeset uk.gov.pay:add_index_on_service_id_to_transaction_table runInTransaction:false
+CREATE INDEX CONCURRENTLY transaction_service_id_idx
+    ON transaction(service_id);
--- a/src/main/resources/migrations/00064_add_service_id_and_live_coumn_to_event_table.sql
+++ b/src/main/resources/migrations/00064_add_service_id_and_live_coumn_to_event_table.sql
@@ -1,0 +1,8 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:add_service_id_and_live_to_event_table
+ALTER TABLE event
+ADD COLUMN service_id VARCHAR(32),
+ADD COLUMN live BOOLEAN;
+
+--rollback ALTER TABLE event DROP COLUMN service_id, DROP column live;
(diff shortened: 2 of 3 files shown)


src/main/resources/migrations/00065_add_service_id_column_to_transaction_table.sql
```diff
@@ -0,0 +1,11 @@
...
+
+--changeset uk.gov.pay:add_service_id_to_transaction_table
+ALTER TABLE transaction
+    ADD COLUMN service_id VARCHAR(32);
+
+--rollback ALTER TABLE transaction DROP COLUMN service_id;
+
+--changeset uk.gov.pay:add_index_on_service_id_and_live_to_transaction_table runInTransaction:false
+CREATE INDEX CONCURRENTLY transaction_service_id_live_idx
+    ON transaction(service_id, live);
```
sfount, line 11: I'm not sure a multi column index will actually be used by the Postgres planner in this case, because live can only have to columns I _think_ it will generally opt to use the service id index where possible and filter out live or non live values. 

As I understand it there's already a partial index on the live column so we can probably get away with just indexing the `service_id` column. 

Happy to chat about this, I wouldn't be 100% without trying out a few different plans.

Maybe @kbottla has thoughts?

sfount (commented):
SQL looks good to me, question about indexes but I think this can go through following clarification

src/main/resources/migrations/00065_add_service_id_column_to_transaction_table.sql
kbottla: I think full index will be used most of the time except for high volume services. Due to low cardinality (service_id and live flag) for high volume services, postgres will likely choose sequential scan for search. This is an existing problem though (for search by gateway account for high volume services or with sparse data).

mrlumbu: Might be right, but I suspect it will probably help. All queries of form `... WHERE service_id = 'blah'...` will be able to use the multicolumn index because `service_id` is the primary sort criterion. It may not then need the secondary bit, but I think for services with large volumes of tx it probably would. How about we do the migration as is and run  some `ANALYZE`s on test DB?

mrlumbu: heh, @kbottla and I said the opposite thing, but came to same conclusion. So I think let's leave as is, and we'll have a story to test ledger performance once all this data is in and populated?

mrlumbu: final final decision, have decided to go with single column index. But we will look at performance of this when possible. Postgres docs say 
>Multicolumn indexes should be used sparingly. In most situations, an index on a single column is sufficient and saves space and time

https://www.postgresql.org/docs/10/indexes-multicolumn.html
so this seems sensible for now.

sfount (approved):

PR #1400, opened 2021-09-09
PP-8461 add pact test for gateway requires 3ds authorisation

SandorArpa:
- add new pact test for GatewayRequires3dsAuthorisation event


kbottla (approved):
LGTM!

PR #1403, opened 2021-09-10
PP-8440 Store and retrieve `service_id` and `live` (redux with bug fix)

gidsg:
Re-introduces changes reverted in https://github.com/alphagov/pay-ledger/pull/1383 with additional fixes to prevent a null pointer exception in the case where `live` is `null` either in the top-level and/or `event_details` of an Event.


alexbishop1 (approved):
@mrlumbu and @alexbishop1 paired with @gidsg on this and are confident it requires no further review in line with our software development guidelines.

PR #1404, opened 2021-09-10
PP-8534 - Include 3ds info for get payment

nimalank7:
Description:
- For the shape of the JSON see the JIRA ticket for PP-8534
- If either requires_3ds or version_3ds is in transaction_details then Ledger will return authorisation_summary JSON blob


SandorArpa (approved):
LGTM

PR #1409, opened 2021-09-14
PP-8543 - Add search transaction authorisation summary test

nimalank7:
Description:
- The previous PR (https://github.com/alphagov/pay-ledger/pull/1404) means that AuthorisationSummary is returned in TransactionView
which is used by TransactionSearchResponse. Hence searching for transactions returns the AuthorisationSummary JSON blob.
- Adds a test for this


SandorArpa (approved):
LOKTM

PR #1418, opened 2021-09-21
[PP-8695] Remove ECS Deploy Step from Jenkinsfile

dj-maisy:
## What?
This should un-break the Jenkins Pact Tests now that microservices_v1 are gone.


katstevens (approved):
And this one looks like the pact tagging is split out into different steps anyway...

PR #1430, opened 2021-09-30
BAU: take the first non-null value for `live`

gidsg:
As live may be null on some events, take the first non-null value rather than the value of the latest event.


gidsg:
Thanks @alexbishop1 I have forced pushed and addressed your comments in ed0ee10.

PR #1433, opened 2021-10-01
BAU Remove randomness from pact tests

stephencdaly:
The pact broker does some clever stuff to compare pact files, so every single version of the consumer doesn't need to be verified against every single version of the provider if the pact hasn't changed.

When there is randomness in the pact this is prevented, meaning apps have to be deployed in a particular order, and only certain versions can be deployed depending on which commit the consumer was verified against.


alexbishop1 (approved):
I assume the green build means there’s nothing relying on the values actually being different for each fixture.

PR #1524, opened 2021-12-10
PP-9010 Upgrade log4j-core and log4j-api to 2.15.0

alexbishop1:
Upgrade log4j-core and log4j-api (which we do not use but some dependencies do) to version 2.15.0 to fix CVE-2021-44228.


danworth (approved):
Looks good 🏅

PR #1533, opened 2021-12-14
PP-9017 Publish incoming events to SNS

mrlumbu:
Adds a new action for ledger on recieving an event.
If SNS is enabled, Ledger publishes the event to AWS SNS,
using the provide topic.

If SNS is not enabled then publish should not be tried
and everything should behave as before.

I have introduced an enum to represent available SNS topics
to rest of app - feels nicer and safer than any of the
alternatives.

The publisher itself is very simple - don't think anything
controversial.

At the moment if any exception in SNS publishing will
be swallowed and only logged. This is to make totally sure
this does not break environments where SNS is not currently
provisioned.

Set up emitter and hook in to event processing life
cycle


.gitignore
```diff
@@ -30,3 +30,5 @@ target
 bin
 
 .DS_Store
+
+.java-version
```
alexbishop1, line 34: I’m inclined to just check this file in.

build-local.sh
```diff
@@ -4,5 +4,5 @@ set -e
 
 cd "$(dirname "$0")"
 
-mvn -DskipITs clean verify
+mvn -DskipTests clean verify
```
alexbishop1, line 7: Are you sure?

pom.xml
```diff
@@ -142,6 +142,11 @@
             <artifactId>aws-java-sdk-sqs</artifactId>
             <version>1.12.129</version>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>sns</artifactId>
+            <version>2.17.46</version>
```
alexbishop1, line 148: Latest is version 2.17.99.

src/main/java/uk/gov/pay/ledger/eventpublisher/EventPublisher.java
```diff
@@ -0,0 +1,35 @@
...
+    public void publishMessageToTopic(String message, TopicName topicName) throws EventPublisherException{
+        var topicArn = topicNameArnMapper.getArnForTopicName(topicName);
+        try {
+            PublishRequest request = PublishRequest.builder()
+                    .message(message)
+                    .topicArn(topicArn)
+                    .build();
+            snsClient.publish(request);
+        } catch (Exception e) {
+            LOGGER.warn("Failed to publish message to SNS");
```
alexbishop1, line 31: We should probably have some indication of what the message is.

src/test/java/uk/gov/pay/ledger/queue/EventMessageHandlerTest.java
```diff
@@ -119,4 +139,28 @@ void shouldScheduleMessageForRetry_WhenEventIsNotProcessedSuccessfully() throws
...
+
+    @Test
+    void shouldPublishToMessageWhenSnsEnabled() throws QueueException {
+        Event event = aQueuePaymentEventFixture().toEntity();
+        when(eventMessage.getEvent()).thenReturn(event);
+
+        eventMessageHandler.handle();
+
+        verifyNoInteractions(eventPublisher);
+    }
```
alexbishop1, line 165: The names of these tests seem to be the wrong way around!

alexbishop1 (changes requested):
Basically fine.

src/main/java/uk/gov/pay/ledger/eventpublisher/EventPublisher.java
mrlumbu: yes, but at this point the message is just a string, which we obvs don't want to log. Could explicitly pass in message id in to method for logging purposes?

mrlumbu: having discussed, agreed that responsibility for logging should be in caller, EventPublisher just needs to throw sufficiently informative exception

alexbishop1 (approved):

PR #1541, opened 2021-12-17
PP-9017 Handle case where topic arn not set

mrlumbu:


src/main/java/uk/gov/pay/ledger/eventpublisher/TopicNameArnMapper.java
```diff
@@ -2,20 +2,19 @@
...
-        this.nameToArnMap  = Map.of(
-                TopicName.CARD_PAYMENT_EVENTS,
-                ledgerConfig.getSnsConfig().getCardPaymentEventsTopicArn()
-        );
+        this.snsConfig  = ledgerConfig.getSnsConfig();
     }
     public String getArnForTopicName(TopicName topicName) {
-        return nameToArnMap.get(topicName);
+        return Map.of(TopicName.CARD_PAYMENT_EVENTS, snsConfig.getCardPaymentEventsTopicArn())
+                .get(topicName);
```
sfount, line 18: It might be helpful if this threw a relevant error or null if no topic ARN was configured in the environment, which would then be handled further up. I don't think it's a deal breaker as this code can only be reached by setting a separate explicit env flag

sfount (approved):

src/main/java/uk/gov/pay/ledger/eventpublisher/TopicNameArnMapper.java
sfount: This also instantiates a map and looks up config on every publish (which will have a significant throughput) so could also be optimised by removing it from each call

PR #1582, opened 2022-01-12
PP-9038 Add fee breakdown to CSV

kbottla:
## WHAT
- Adds fee breakdown fields to CSV download if the query param `fee_breakdown_headers` is set.
- Fields appear as below in CSV
image


SandorArpa (approved):
LGTM

PR #1599, opened 2022-01-19
PP-8951 update the CSV transaction factory to zero `net` when appropriate

nlsteers:
### WHAT
When downloading transaction details in CSV format, Ledger now returns a `0.00` net value for a failed payment that has no `net_amount`. Refund behaviour is unchanged.

README tables reformatted with correct width.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
@@ -10,6 +10,7 @@ import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
 import uk.gov.pay.ledger.transaction.state.ExternalTransactionState;
 import uk.gov.pay.ledger.transaction.state.PaymentState;
 import uk.gov.pay.ledger.transaction.state.RefundState;
+import uk.gov.pay.ledger.transaction.state.TransactionState;
 
 import java.io.IOException;
 import java.text.DecimalFormat;
@@ -64,7 +65,7 @@ public class CsvTransactionFactory {
     private static final String FIELD_MOTO = "MOTO";
     private static final String FIELD_PAYMENT_PROVIDER = "Payment Provider";
     private static final String FIELD_3D_SECURE_REQUIRED = "3-D Secure Required";
-    private ObjectMapper objectMapper;
+    private final ObjectMapper objectMapper;
 
     @Inject
     public CsvTransactionFactory(ObjectMapper objectMapper) {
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
```diff
@@ -95,8 +96,12 @@ public CsvTransactionFactory(ObjectMapper objectMapper) {
                 result.put(FIELD_GOVUK_PAYMENT_ID, transactionEntity.getExternalId());
                 result.put(FIELD_AMOUNT, penceToCurrency(transactionEntity.getAmount()));
                 result.put(FIELD_TOTAL_AMOUNT, penceToCurrency(totalAmount));
-                result.put(FIELD_NET, penceToCurrency(netAmount));
                 result.put(FIELD_FEE, penceToCurrency(transactionEntity.getFee()));
+                if (nonNull(transactionEntity.getFee())) {
```
kbottla, line 100: We only need to set net_amount to `0` for failed transactions.

Fee is only applicable for stripe payments. So for worldpay payments that are successful this will result in net_amount of `0` for the csv.

stephencdaly: I guess for any payment that doesn't have a `net_amount` set on the transaction, we should default to 0. We also want to return 0 for in progress payments too I think, not just failed payments.

Services would only see the Net column for Worldpay payments in the CSV if they have a mix of Worldpay and Stripe services, and download from the "View transactions for all services" screen. Or I guess if they've switched provider from Worldpay to Stripe (or the other way round) and are looking at historic transactions.
Do we want to check the payment provider in here to only return 0 for Stripe, otherwise, return the amount?

nlsteers: Updated to match the above requirements.

src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
```diff
@@ -71,40 +73,95 @@ public void toMapShouldReturnMapWithCorrectCsvDataForPaymentTransaction() {
...
 
-        Map<String, Object> csvDataMap = csvTransactionFactory.toMap(refundTransactionFixture.toEntity());
+        Map<String, Object> csvDataMap = csvTransactionFactory.toMap(transactionEntity);
 
-        assertPaymentDetails(csvDataMap, refundTransactionFixture.toEntity());
-        assertThat(csvDataMap.get("Amount"), is("-0.99"));
-        assertThat(csvDataMap.get("Provider ID"), is(refundTransactionFixture.getGatewayTransactionId()));
-        assertThat(csvDataMap.get("GOV.UK Payment ID"), is(refundTransactionFixture.getParentExternalId()));
+        assertPaymentDetails(csvDataMap, transactionEntity);
+        assertThat(csvDataMap.get("Amount"), is("-1.00"));
```
kbottla, line 94: Is this default for the amount set somewhere ? `Total Amount` is `-1.23` and the amount is `-1.00` but for refund these should be the same. Guessing this is because of test setup ?

src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
```diff
@@ -71,40 +73,95 @@ public void toMapShouldReturnMapWithCorrectCsvDataForPaymentTransaction() {
...
+        assertThat(csvDataMap.get("Total Amount"), is("-1.23"));
+        assertThat(csvDataMap.get("Net"), is("-1.23"));
         assertThat(csvDataMap.get("MOTO"), is(nullValue()));
     }
 
     @Test
-    public void toMapShouldIncludeExternalMetadataFields() {
+    void toMapShouldIncludeNetForFailedTransactionWhenAvailable() {
+
+        TransactionEntity transactionEntity = transactionFixture.withNetAmount(-5L).toEntity();
```
kbottla, line 112: Can we set a failed state explicitly ? Default is SUBMITTED and changing it would fail this test.

kbottla: And another test where `Net` is set to `0` if the transaction state is other than SUCCESS ?

nlsteers: > And another test where `Net` is set to `0` if the transaction state is other than SUCCESS ?

There is a test that covers this scenario already, the default fixture is in a failure state `toMapShouldReturnMapWithCorrectCsvDataForPaymentTransaction`.

src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
nlsteers: Yes, the fixture in the setup method sets `amount` to `100L`. I'll clarify the value in the test case.

kbottla (approved):
LGTM!

PR #1664, opened 2022-02-21
PP-9252 Add index on gateway_account_id and created_date columns

kbottla:
## WHAT
- Adds a multicolumn index on ‘gateway_account and created_date’ columns of transaction table which could address slow query issues (which also fixed scheduled performance job for search https://build.ci.pymnt.uk/view/HUD/job/run-scheduled-performance-test-for-search/208/ that has been failing for a long time) when services load all services transactions page.
- This may not address the slow queries issue fully because of the different combination of search params allowed on the transaction table. Need to look into different indexes and remove any not required separately.


SandorArpa (approved):
LGTM

PR #1668, opened 2022-02-22
PP-9311 Create gateway_account_metadata table

kbottla:
## WHAT
- Add new table `gateway_account_metadata` which stores gateway accounts and the metadata keys accounts used for payments overtime. This is to be used to get CSV metadata headers.


stephencdaly (approved):
LGTM

SandorArpa (approved):
LOKTM

PR #1674, opened 2022-02-23
PP-9311 Update CSV service to use gateway account metadata table

stephencdaly:
Use the gateway_account_metadata table to get all metadata keys for
gateway accounts included in a CSV download, rather than just the keys
for transactions that are included.

We've changed this behaviour as querying with joins on the transaction
table to get the keys was very expensive and was causing CSV downloads
to take a long time or time out.

Co-authored-by: Stephen Daly <stephen.daly@digital.cabinet-office.gov.uk>


nlsteers:
I’m confident that this change does not require further review and meets the secure coding requirements of GOV.UK Pay.

nlsteers (approved):

PR #1675, opened 2022-02-23
PP-9258 add new resource type

SandorArpa:
## WHAT
- Making sure ledger is able to process the new resource type without throwing an error, but just ignore the event for now.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/queue/eventprocessor/DisputeEventProcessor.java
+++ b/src/main/java/uk/gov/pay/ledger/queue/eventprocessor/DisputeEventProcessor.java
@@ -1,0 +1,18 @@
+package uk.gov.pay.ledger.queue.eventprocessor;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.ledger.event.model.Event;
+
+public class DisputeEventProcessor extends EventProcessor{
+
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/ledger/queue/eventprocessor/DisputeEventProcessor.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.pay.ledger.queue.eventprocessor;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import uk.gov.pay.ledger.event.model.Event;
+
+public class DisputeEventProcessor extends EventProcessor{
```
stephencdaly, line 7: We don't strictly need this yet. We're not going to be doing any processing of these kind of events until the second half of the chargebacks work, so this will just sit around logging. No harm in adding it though.

stephencdaly (approved):
LGTM!

src/main/java/uk/gov/pay/ledger/queue/eventprocessor/DisputeEventProcessor.java
SandorArpa: I'm aware, I just thought this looks better than some // TODO 😃

PR #1678, opened 2022-02-24
PP-9311 Clean up unused DAO queries

nlsteers:
We no longer look up metadata keys on individual transactions as part of Ledger performance improvement work


stephencdaly (approved):
LGTM!

PR #1679, opened 2022-02-24
BAU Update images used by testscontainers for M1 compatibility

nlsteers:
Updated the following images used by `testcontainers`:
-   `testcontainersofficial/ryuk` -> `testcontainers/ryuk:0.3.3`
-  `oribio16/alpine-sqs` -> `mvisonneau/alpine-sqs:1.2.0`

### DIFF
--- a/src/test/resources/testcontainers.properties
+++ b/src/test/resources/testcontainers.properties
@@ -1,2 +1,2 @@
-# Digest of image testcontainersofficial/ryuk:latest for linux/amd64 from tags list on https://hub.docker.com/r/testcontainersofficial/ryuk
-ryuk.container.image=testcontainersofficial/ryuk@sha256:8fdd715bf34975d088273bcea497fe048f9b768ba4c7fc690ab6e4ddc995e317
+# arm64/amd64 friendly ryuk image
+ryuk.container.image=testcontainers/ryuk:0.3.3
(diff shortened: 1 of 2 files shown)


katstevens:
Am I right that this image is only used in the pact tests?

nlsteers:
These images are used by any tests that require `testcontainers`, e.g. anything that needs to test Postgres / SQS interactions.

src/test/resources/testcontainers.properties
```diff
@@ -1,2 +1,2 @@
-# Digest of image testcontainersofficial/ryuk:latest for linux/amd64 from tags list on https://hub.docker.com/r/testcontainersofficial/ryuk
-ryuk.container.image=testcontainersofficial/ryuk@sha256:8fdd715bf34975d088273bcea497fe048f9b768ba4c7fc690ab6e4ddc995e317
+# arm64/amd64 friendly ryuk image
+ryuk.container.image=testcontainers/ryuk:0.3.3
```
katstevens, line 2: Is there any difference between `testcontainers` and `testcontainersofficial`? They both seem up to date 🤔

nlsteers: afaict, the difference is that the former vends multi platform images whereas the latter does not...

katstevens (approved):
OK, let's try this out 👍

PR #1742, opened 2022-03-21
PP-9278 Fix serialisation of SNS message

stephencdaly:
When serialising the event to send to SNS, use the EventMessageDto,
rather then the internal Event model. This ensures that the field names
are unchanged from the event received from connector. In particular,
the "event_details" fields is renamed to "event_data" in the internal
ledger Event model.

This field isn't deserialised by webhooks, so we won't break anything
by changing it.


SandorArpa (approved):
LGTM

PR #1747, opened 2022-03-22
BAU Serialise event date using standard pay format

sfount:
Rename `event_date` to `timestamp` match the original event spec (from producer
Connector). The event store shouldn't be doing any renaming on event
fields before publishing.

Downstream services receiving events parse event metadata. Now that this
is the publishing DTO, make sure a standard format is used for time.


mrlumbu (approved):
Makes things less confusing, and nothing in prod currently relies on this

PR #1754, opened 2022-03-24
PP-9281 Derive service_id from events

kbottla:
## WHAT

- Derived service_id if any one event in the given list has service_id.
- This is because some events from connector doesn't have service_id set and service_id on the transaction could be null if taken from latest event.


stephencdaly (approved):
LGTM

PR #1779, opened 2022-04-04
PP-9456 OpenAPI specs for transaction endpoints

kbottla:
## WHAT
- Updates to generate openAPI specs for transaction endpoints
- CSV download endpoint is hidden as openAPI specs does not support multiple paths with different headers
- TransactionSearchParams - moved JAX-RS annotations and added openAPI specs to setter methods as swagger library is not generating all query params if annotations are on both fields and methods.


stephencdaly (approved):
LGTM

PR #1852, opened 2022-05-04
PP-9531 Agreements and payment instrument DAO

sfount:
Entities to be mapped to agreement and payment instrument table rows.

DAO and JDBI row mappers for agreements and payment instruments.

Fetch agreements with their latest payment instrument using Postgres
`SELECT DISTINCT ON`.

Test coverage to assert that row mappers are correctly mapping and
casting fields.

Test coverage to assert that the latest payment instrument (by created
date) is associated with an agreement when pulled out of the database
and mapped to an agreement entity.


alexbishop1 (approved):
I’d prefer the `ZonedDateTime`s to be `Instant`s but it matches the rest of ledger, so…

PR #1858, opened 2022-05-05
PP-9526 Sync event write endpoint

sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/1857

---

Sync write event endpoint. Accepts an array of standard Pay event JSON.

This interface directly matches events that are consumed from SQS.

Add coverage to assert events passed into this list (POST body) are
written to the database.

Ensure that any interactions with `jdbi` are rolled back if
any
individual event fails to process, this includes:
* writing events
* domain object projection

Add coverage with an incomplete event (missing resource external ID) --
this will fail a nullable column constraint and cause the batch to rollback.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/app/LedgerModule.java
+++ b/src/main/java/uk/gov/pay/ledger/app/LedgerModule.java
@@ -44,6 +44,7 @@ public class LedgerModule extends AbstractModule {
     protected void configure() {
         bind(LedgerConfig.class).toInstance(configuration);
         bind(Environment.class).toInstance(environment);
+        bind(Jdbi.class).toInstance(jdbi);
     }
 
     @Provides
--- a/src/main/java/uk/gov/pay/ledger/event/resource/EventResource.java
+++ b/src/main/java/uk/gov/pay/ledger/event/resource/EventResource.java
@@ -15,9 +15,14 @@ import uk.gov.pay.ledger.event.dao.EventDao;
 import uk.gov.pay.ledger.event.model.Event;
 import uk.gov.pay.ledger.event.model.EventTicker;
 import uk.gov.pay.ledger.exception.ErrorResponse;
+import uk.gov.pay.ledger.queue.EventMessage;
+import uk.gov.pay.ledger.queue.EventMessageDto;
+import uk.gov.pay.ledger.queue.EventMessageHandler;
+import uk.gov.service.payments.commons.queue.exception.QueueException;
 
 import javax.validation.constraints.NotEmpty;
 import javax.ws.rs.GET;
+import javax.ws.rs.POST;
 import javax.ws.rs.Path;
 import javax.ws.rs.PathParam;
 import javax.ws.rs.Produces;
@@ -26,6 +31,7 @@ import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.Response;
 import java.time.ZonedDateTime;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
@@ -36,10 +42,12 @@ public class EventResource {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(EventResource.class);
     private final EventDao eventDao;
+    private final EventMessageHandler eventMessageHandler;
 
     @Inject
-    public EventResource(EventDao eventDao) {
+    public EventResource(EventDao eventDao, EventMessageHandler eventMessageHandler) {
         this.eventDao = eventDao;
+        this.eventMessageHandler = eventMessageHandler;
     }
 
     @Path("/{eventId}")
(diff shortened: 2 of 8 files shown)


src/main/java/uk/gov/pay/ledger/app/LedgerModule.java
```diff
@@ -156,4 +157,4 @@ public SqsQueueService provideSqsQueueService(AmazonSQS amazonSQS, LedgerConfig
                 ledgerConfig.getSqsConfig().getMessageMaximumWaitTimeInSeconds(),
                 ledgerConfig.getSqsConfig().getMessageMaximumBatchSize());
     }
-}
+}
```
alexbishop1, line 160: Dude, your IDE is at war with the Unix convention that text files end with a newline.

src/main/java/uk/gov/pay/ledger/event/resource/EventResource.java
```diff
@@ -58,6 +66,22 @@ public Event getEvent(@PathParam("eventId") Long eventId) {
...
     }
 
+    @POST
+    @Timed
+    public Response writeEvent(List<EventMessageDto> events) throws QueueException {
+        try {
+            eventMessageHandler.processEventBatch(events
+                    .stream()
+                    .map(eventMessageDto -> EventMessage.of(eventMessageDto, null))
+                    .collect(Collectors.toList())
```
alexbishop1, line 76: Please use the (annoyingly verbose) `Collectors.toUnmodifiableList()` if it’s not going to be modified.

src/main/java/uk/gov/pay/ledger/event/resource/EventResource.java
```diff
@@ -58,6 +66,22 @@ public Event getEvent(@PathParam("eventId") Long eventId) {
...
+            );
+        } catch (Exception ignored) {
```
alexbishop1, line 78: Can we be more specific in terms of what we catch?

src/main/java/uk/gov/pay/ledger/queue/EventMessage.java
```diff
@@ -17,7 +19,9 @@ public static EventMessage of(EventMessageDto eventDto, QueueMessage queueMessag
     }
 
     public String getQueueMessageId() {
-        return queueMessage.getMessageId();
+        return Optional.ofNullable(queueMessage)
+                .map(QueueMessage::getMessageId)
+                .orElse(null);
```
alexbishop1, line 24: How much work would it be make this method return an optional?

src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
```diff
@@ -70,6 +74,15 @@ public void handle() throws QueueException {
         }
     }
 
+    // provides a transactional guarantee, if any of the events fail to process, none of the events will be persisted
+    public void processEventBatch(List<EventMessage> messages) throws QueueException {
+       jdbi.useTransaction(handle -> {
+            for (EventMessage message : messages) {
+                processSingleMessage(message);
+            }
+        });
```
alexbishop1, line 83: Or `jdbi.useTransaction(handle -> messages.forEach(this::processSingleMessage));` 😀

src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
```diff
@@ -115,7 +128,9 @@ private void processSingleMessage(EventMessage message) throws QueueException {
 
         if (response.isSuccessful()) {
             eventDigestHandler.processEvent(event, response.getState() == INSERTED);
-            eventQueue.markMessageAsProcessed(message);
+            if (message.getQueueMessageReceiptHandle() != null) {
+                eventQueue.markMessageAsProcessed(message);
+            }
```
alexbishop1, line 133: So I’m more persuaded about the optional thing now.

src/main/resources/migrations/00072_remove_event_sqs_message_id_null_constraint.sql
```diff
@@ -0,0 +1,4 @@
+--liquibase formatted sql
+
+--changeset uk.gov.pay:remove_event_sqs_message_id_not_null_constraint
+ALTER TABLE event ALTER COLUMN sqs_message_id DROP NOT NULL;
```
alexbishop1, line 4: This has sneaked in from the other PR!

alexbishop1 (changes requested):
Looks basically fine.

src/main/resources/migrations/00072_remove_event_sqs_message_id_null_constraint.sql
sfount: It will go when rebased! Just borrows the commit for integration tests

src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
sfount: I like this a lot but unfortunately the method specifies a thrown exception and java wants you to deal with it. Briefly looking around either you have to do some interface hackery or the scope of the current expression just has it thrown for the scope around the callback.

alexbishop1: Exceptions are the enemy of lambdas.

src/main/java/uk/gov/pay/ledger/queue/EventMessageHandler.java
```diff
@@ -115,7 +128,9 @@ private void processSingleMessage(EventMessage message) throws QueueException {
 
         if (response.isSuccessful()) {
             eventDigestHandler.processEvent(event, response.getState() == INSERTED);
-            eventQueue.markMessageAsProcessed(message);
+            if (message.getQueueMessageReceiptHandle().isPresent()) {
+                eventQueue.markMessageAsProcessed(message);
+            }
```
alexbishop1, line 133: Is an exception woe stopping us from using `message.getQueueMessageReceiptHandle.ifPresent(receiptHandle -> eventQueue.markMessageAsProcessed(message));`?

sfount: spot on, same issue as above

src/main/java/uk/gov/pay/ledger/app/LedgerModule.java
sfount: Thanks for flagging, found and unset the rogue setting, no ideological beef here

PR #1861, opened 2022-05-06
PP-9533 Agreement resource

sfount:
Add endpoints to
* get one agreement
* search for agreements

Agreement search supports standard Pay result response
* pagination
* filtering on `status` and `reference`

Lifts core pagination and search string templating methodology directly from transaction search.

Either `gateway_account_id` or `service_id` is required to access the
search endpoint. This should guard against consumers accidentally
requesting data across services.

Adds agreement service to manage search functionality.

HTTP endpoints generally follow a pattern of serialising an external
facing model rather than using the database entities. Add agreement and
payment instrument shaped models.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementSearchParams.java
@@ -24,7 +24,7 @@ public class AgreementSearchParams extends SearchParams {
     private static final String REFERENCE_FIELD = "reference";
     private static final long DEFAULT_PAGE_NUMBER = 1L;
     private static final long DEFAULT_MAX_DISPLAY_SIZE = 20L;
-    private static final long DEFAULT_DISPLAY_SIZE = 20L;
+    public static final long DEFAULT_DISPLAY_SIZE = 20L;
 
     private long maxDisplaySize = DEFAULT_MAX_DISPLAY_SIZE;
 
@@ -133,7 +133,7 @@ public class AgreementSearchParams extends SearchParams {
                 queryMap.put(LIVE_FIELD, live);
             }
         }
-        return queryMap;
+        return Map.copyOf(queryMap);
     }
 
     public List<String> getGatewayAccountIds() {
--- a/src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementSearchParams.java
@@ -1,0 +1,208 @@
+package uk.gov.pay.ledger.agreement.resource;
+
+import uk.gov.pay.ledger.common.search.SearchParams;
+
+import javax.validation.constraints.AssertTrue;
+import javax.validation.constraints.NotNull;
+import javax.ws.rs.DefaultValue;
+import javax.ws.rs.QueryParam;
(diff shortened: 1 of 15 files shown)


src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementSearchParams.java
```diff
@@ -0,0 +1,208 @@
...
+                } else {
+                    queryMap.put(REFERENCE_FIELD, likeClause(reference));
+                }
+            }
+
+            if (live != null) {
+                queryMap.put(LIVE_FIELD, live);
+            }
+        }
+        return queryMap;
```
alexbishop1, line 136: Not `Map.copyOf(queryMap)`?

src/main/java/uk/gov/pay/ledger/agreement/service/AgreementService.java
```diff
@@ -0,0 +1,54 @@
...
+    public Optional<Agreement> findAgreement(String externalId) {
+        return agreementDao.findByExternalId(externalId)
+                .map(Agreement::from);
+    }
+
+    public AgreementSearchResponse searchAgreements(AgreementSearchParams searchParams, UriInfo uriInfo) {
+        var agreements = agreementDao.searchAgreements(searchParams)
+                .stream()
+                .map(Agreement::from)
+                .collect(Collectors.toList());
```
alexbishop1, line 34: `toUnmodifiableList()` please.

src/main/java/uk/gov/pay/ledger/agreement/service/AgreementService.java
```diff
@@ -0,0 +1,54 @@
...
+        var total = agreementDao.getTotalForSearch(searchParams);
+
+        long size = searchParams.getDisplaySize();
+        if (total > 0 && searchParams.getDisplaySize() > 0) {
+            long lastPage = (total + size - 1) / size;
+            if (searchParams.getPageNumber() > lastPage || searchParams.getPageNumber() < 1) {
```
alexbishop1, line 40: I’m assuming this is copied from somewhere else.

src/test/java/uk/gov/pay/ledger/agreement/resource/AgreementResourceIT.java
```diff
@@ -0,0 +1,117 @@
...
+                .statusCode(Response.Status.OK.getStatusCode())
+                .contentType(JSON)
+                .body("total", is(0))
+                .body("count", is(0))
+                .body("page", is(1))
+                .body("results.size()", is(0));
+    }
+
+    @Test
+    public void shouldSearchWithPaginationForAgreements() {
```
alexbishop1, line 77: I’d be more convinced if this pagination test involved more than one page!

alexbishop1 (changes requested):
Looks basically fine.

src/main/java/uk/gov/pay/ledger/agreement/service/AgreementService.java
sfount: Yeah that's correct, it follows the exact same pattern as transaction search which is also used by the admin tool and public API. Maybe it's worth extracting this as a simple util as a following cleanup step

alexbishop1 (approved):

PR #1864, opened 2022-05-09
PP-9534 Project agreements/ payment instruments

sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/1861 (in order to extend `AgreementService`)

---

Add entity factory to reduce event streams to `AgreementEntity` and
`PaymentInstrumentEntity`.

Use the standard ledger reduce -> object mapper for standard fields.
Specifically extract top level event digest fields to set:
* resource id
* created date
* service id
* live
* event count

Hook agreement and payment instrument event stream processing into the
`EventDigestHandler`.

This will allow events consumed from SQS or passed to the event endpoint
to be processed with resource type `AGREEMENT` and `PAYMENT_INSTRUMENT`.

Update tests to accomidate now supporting agreements, cover individual
entity mapping and end to end (integration) consuming and projecting
events.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntityFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntityFactory.java
@@ -1,0 +1,35 @@
+package uk.gov.pay.ledger.agreement.entity;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.inject.Inject;
+import uk.gov.pay.ledger.event.model.EventDigest;
+
+public class AgreementEntityFactory {
+
+    private final ObjectMapper objectMapper;
+
+    @Inject
+    public AgreementEntityFactory(ObjectMapper objectMapper){
+        this.objectMapper = objectMapper;
+    }
+
+    public AgreementEntity create(EventDigest eventDigest) {
+        var entity = objectMapper.convertValue(eventDigest.getEventAggregate(), AgreementEntity.class);
+
+        entity.setExternalId(eventDigest.getResourceExternalId());
+        entity.setCreatedDate(eventDigest.getEventCreatedDate());
+        entity.setServiceId(eventDigest.getServiceId());
+        entity.setLive(eventDigest.isLive());
+        entity.setEventCount(eventDigest.getEventCount());
+        return entity;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntityFactory.java
```diff
@@ -0,0 +1,35 @@
...
+
+        entity.setExternalId(eventDigest.getResourceExternalId());
+        entity.setCreatedDate(eventDigest.getEventCreatedDate());
+        entity.setServiceId(eventDigest.getServiceId());
+        entity.setLive(eventDigest.isLive());
+        entity.setEventCount(eventDigest.getEventCount());
+        return entity;
+    }
+
+    public PaymentInstrumentEntity createPaymentInstrument(EventDigest eventDigest) {
```
alexbishop1, line 27: Feels a bit odd that the `AgreementEntityFactory` makes `PaymentInstruments`.

src/test/java/uk/gov/pay/ledger/queue/QueueMessageReceiverIT.java
```diff
@@ -359,4 +362,99 @@ public void shouldHandleDisputeTypeEvent() throws InterruptedException, JsonProc
...
+        Thread.sleep(500);
+
+        given().port(rule.getAppRule().getLocalPort())
+                .contentType(JSON)
+                .get("/v1/agreement/" + agreementId)
+                .then()
+                .statusCode(200)
+                .body("external_id", is(agreementId))
+                .body("payment_instrument.external_id", is("a-valid-instrument-id"))
+                .body("payment_instrument.card_details.cardholder_name", is("A paying user name"));
```
alexbishop1, line 458: Feels like we could check the other properties too.

alexbishop1 (dismissed):
Looks good. A few suggestions but none of them showstoppers.

sfount:
@alexbishop1 totally agreed with comments, could come later but have followed up now to make sure they get in https://github.com/alphagov/pay-ledger/pull/1864/commits/bf09669b1f4542f30024e982c736a9b4da939b74

alexbishop1 (approved):

PR #1991, opened 2022-07-07
PP-9626: project dispute events

alan-gds:
* Repurpose RefundEventProcessor to apply to both refunds and disputes - rename to ChildTransactionEventProcessor. Events for both resource types are projected in the same way, pulling in the payment information for certain fields.
* When processing a dispute event, call the ChildTransactionEventProcessor, which will project the dispute transaction including the parent payment information.
* Added transaction states for disputes and mapped them to the SalientEventTypes.
* Ensure that transaction search will ignore disputes until we've added a response mapping for them.

Other bits:
* Added a static set PAYMENT_TRANSACTION_STATES to TransactionState containing the states that apply to payments. This is required as it is used to construct a report of payments by state, which we don't want the dispute states to be included in.

Tests:
* Added an integration test to test projection of dispute events.
* Removed existing test from EventDigestHandlerTest for disputes, but didn't replace it. This test class is a bit of a strange. integration/unit test hybrid and will be removed in a follow-up commit as it duplicates tests that should be in the unit tests for the `EventProcessor` implementations. The logic in EventDigestHandler itself is tested by integration tests.
* Added a test to ensure that disputes are ignored by the search transactions endpoint for the time being.


SandorArpa (approved):
LGTM

PR #1994, opened 2022-07-08
PP-9627 implement get one dispute

SandorArpa:
- implementing getting a dispute by its externalId
- add Dispute model
- extend TransactionView.from method to handle Dispute
- add a mapper for dispute reasons
- add relevant tests

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/Dispute.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/Dispute.java
@@ -1,0 +1,203 @@
+package uk.gov.pay.ledger.transaction.model;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/Dispute.java
```diff
@@ -0,0 +1,215 @@
...
+
+    public Payment getPaymentDetails() {
+        return paymentDetails;
+    }
+
+    public String getParentTransactionId() {
+        return parentTransactionId;
+    }
+
+    public String getTransactionId() {
```
stephencdaly, line 86: It doesn't look like this is used anywhere. We're already setting the `externalId` field on the base Transaction, so I think this field is unnecessary?

stephencdaly (changes requested):
Looks good - just spotted that there is an unused field

stephencdaly (approved):
LGTM!

PR #1998, opened 2022-07-11
PP-9626 Refactor EventDigestHandlerTest

stephencdaly:
This test was kind of a hybrid integration/unit test, which was performing some unit testing of the EventProcessor implementations.

Move all unit testing of the EventProcessors to dedicated unit tests.

Make EventDigestHandlerTest a pure unit test, so that it is only testing that the correct EventProcessor is selected.


SandorArpa (approved):
LOKTM

PR #2016, opened 2022-07-18
PP-9788 Update DisputeCreated pact test

kbottla:
## WHAT
- Removed `fee` from DisputeCreated fixture as Connector will be sending `fee` only for DISPUTE_LOST event
- Removed `evidence_due_date` from DisputeCreated event fixture. To be added back once connector starts sending evidence_due_date value as a `date` instead of timestamp.


SandorArpa (approved):
LGTM

PR #2036, opened 2022-07-22
PP-9671 Return disputed field on transactions

stephencdaly:
For payment transactions, return a "disputed" field on the transaction in the response to the request GET a single transaction or search transactions.

The value of this field comes from the value of the "disputed" field on the "transaction_details" from the database if this field exists, otherwise it defaults to false.


SandorArpa (approved):
LGTM

PR #2044, opened 2022-07-25
PP-9545 X-Consistent support for GET agreements

sfount:
Adds branching logic to the get agreement resource to optionally check
if the current event stream has more events than the most recently
projected resource.

If `X-Consistent` is set to true, fetch the event digest for that
resource. If there are more events in the event digest than the
projection snapshots event count, return a new projection for the event
digest, otherwise return the existing projection.

If `X-Consistent` is set to false or the header has not been set,
directly return the existing projection.


src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementResource.java
```diff
@@ -44,8 +49,31 @@ public AgreementSearchResponse search(@Valid @BeanParam AgreementSearchParams qu
...
-                .orElseThrow(() -> new WebApplicationException(Response.Status.NOT_FOUND));
+        if (isConsistent != null && isConsistent) {
+            try {
+                var eventDigest = eventService.getEventDigestForResource(agreementExternalId);
+
+                // compare the current event stream to the existing projection snapshot
+                return agreementService.findAgreementEntity(agreementExternalId)
+                        .map(agreementEntity -> {
+                            if (agreementEntity.getEventCount() < eventDigest.getEventCount()) {
+                                // the projection from the database is behind, rebuild a projection using the latest events
```
sfount, line 62: Suggest we have a standard log pattern (maybe extract to an overriden private method) to show the following for every request:
- was consistent requested
- did we rebuild from the event stream or serve the projection

I'd expect these numbers will be useful for tuning this events pattern.

src/test/java/uk/gov/pay/ledger/agreement/resource/AgreementResourceTest.java
```diff
@@ -63,4 +82,121 @@ public void findShouldReturn404_IfMissing() {
...
+        agreementEntity.setEventCount(eventCount);
+        return agreementEntity;
+    }
+
+    private EventDigest stubEventDigest(String agreementId, Integer eventCount) {
+        List<Event> events = new ArrayList<>();
+        for (int i = 0; i < eventCount; i++) {
+            events.add(anEventFixture().withResourceExternalId(agreementId).toEntity());
+        }
+        return EventDigest.fromEventList(events);
```
alexbishop1, line 200: Could also do this for bonus functional points!

```java
return EventDigest.fromEventList(IntStream.range(0, eventCount)
        .mapToObj(i -> anEventFixture().withResourceExternalId(agreementId).toEntity())
        .collect(toUnmodifiableList()));
```

alexbishop1 (changes requested):
Looks good. A shame the different tests end up split across two different files. One suggestion about simplifying some of the logic; the rest is optional.

src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementResource.java
```diff
@@ -44,8 +49,31 @@ public AgreementSearchResponse search(@Valid @BeanParam AgreementSearchParams qu
...
+                                return Agreement.from(agreementService.projectAgreement(eventDigest));
+                            } else {
+                                return Agreement.from(agreementEntity);
+                            }
+                        })
+                        .orElseGet(() -> Agreement.from(agreementService.projectAgreement(eventDigest)));
```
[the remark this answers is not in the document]
alexbishop1: Messed up the filter! Should be `agreementEntity.getEventCount() == eventDigest.getEventCount()`.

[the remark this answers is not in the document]
sfount: I like this, with the named boolean comparison I think it's quite clear what the operation is

src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementResource.java
```diff
@@ -40,12 +47,32 @@ public AgreementSearchResponse search(@Valid @BeanParam AgreementSearchParams qu
...
+                         @HeaderParam(HEADER_PARAM_X_CONSISTENT) Boolean isConsistent,
                          @Context UriInfo uriInfo) {
-        return agreementService.findAgreement(agreementExternalId)
-                .orElseThrow(() -> new WebApplicationException(Response.Status.NOT_FOUND));
+        if (Boolean.TRUE.equals(isConsistent)) {
+            try {
+                var eventDigest = eventService.getEventDigestForResource(agreementExternalId);
+
+                return agreementService.findAgreementEntity(agreementExternalId)
+                        .filter(agreementEntity -> !databaseProjectionSnapshotIsBehindEventStream(agreementEntity, eventDigest))
```
alexbishop1, line 65: The negation of this conditional make it a bit hard to follow. Either:
- change the filter method so the logic is reversed  (e.g. `databaseProjectionSnapshotIsUpToDateWithEventStream(…)`)
- keep it as `databaseProjectionSnapshotIsBehindEventStream(…)` but don’t negate it and swap the `.map(Agreement::from)` and `Agreement.from(agreementService.projectAgreement(eventDigest)))` bits

alexbishop1 (changes requested):
Thanks for making the changes. Just one thing about the comprehensibility of the boolean logic.

alexbishop1 (approved):

PR #2049, opened 2022-07-26
PP-9545 Add missing payment instrument properties

sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/2048

Adds `type`, `card_type` and `first_digits_card_number` properties to payment instrument entities and mappers.

---

Include the commons enum `PaymentInstrumentType` across the various
entities and mappers required to set and read from the payment
instrument table.

---

Now that the agreement status is represented with a pay java commons
enum, use this when storing projections from the event stream and
reading from the projection database.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntity.java
+++ b/src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntity.java
@@ -3,6 +3,7 @@ package uk.gov.pay.ledger.agreement.entity;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.databind.PropertyNamingStrategies;
 import com.fasterxml.jackson.databind.annotation.JsonNaming;
+import uk.gov.service.payments.commons.model.agreement.AgreementStatus;
 
 import java.time.ZonedDateTime;
 
@@ -14,7 +15,7 @@ public class AgreementEntity {
     private String serviceId;
     private String reference;
     private String description;
-    private String status;
+    private AgreementStatus status;
     private Boolean live;
     private ZonedDateTime createdDate;
     private Integer eventCount;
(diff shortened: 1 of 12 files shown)


src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementEntity.java
```diff
@@ -14,7 +15,7 @@ public class AgreementEntity {
     private String serviceId;
     private String reference;
     private String description;
-    private String status;
+    private AgreementStatus status;
```
alexbishop1, line 18: 🎉

src/test/java/uk/gov/pay/ledger/agreement/resource/AgreementResourceIT.java
```diff
@@ -56,7 +59,10 @@ public void shouldGetAgreementWithPaymentInstrument() {
                 .statusCode(Response.Status.OK.getStatusCode())
                 .contentType(JSON)
                 .body("external_id", is(agreementFixture.getExternalId()))
-                .body("payment_instrument.external_id", is(paymentInstrumentFixture.getExternalId()));
+                .body("payment_instrument.external_id", is(paymentInstrumentFixture.getExternalId()))
+                .body("payment_instrument.type", is(PaymentInstrumentType.CARD.name()))
+                .body("payment_instrument.card_details.card_type", is(CardType.CREDIT.name().toLowerCase()))
```
alexbishop1, line 64: Really this should probably use `.toLowerCase(Locale.ENGLISH)` or `.toLowerCase(Locale.ROOT)` rather than relying on the default locale.

alexbishop1 (approved):

src/test/java/uk/gov/pay/ledger/agreement/resource/AgreementResourceIT.java
sfount: Good catch, we've got a few follow up PRs so will include this in those

PR #2058, opened 2022-07-27
PP-9662 Update pact state

kbottla:
## WHAT
- Updated `dispute lost transaction exists` pact state to set defaults for state params so that the same state can be used by the selfservice pact test


SandorArpa (approved):
LGTM

PR #2067, opened 2022-08-01
PP-9545 Make search param consistent with transactions

sfount:
Transactions expects an `account_id` for search, this is generally
provided by public api. Agreements originally specified this as
`gateway_account_id` as it's more descriptive but change it back to
`account_id` to bring it in line with the other endpoints public api
interacts with.


oswaldquek (approved):

sfount:
Thank @oswaldquek @alexbishop1 can 2i tomorrow

PR #2081, opened 2022-08-05
PP-9829 add webhooks to provider contract test suite

nlsteers:
### WHAT

- added `webhooks` to the provider contract test suite
- moved event based pact tests to a more appropriate package


src/test/java/uk/gov/pay/ledger/pact/ConsumerContractTestSuite.java
```diff
@@ -5,6 +5,7 @@
 import junit.framework.TestSuite;
 import org.junit.runner.RunWith;
 import org.junit.runners.AllTests;
+import uk.gov.pay.ledger.pact.event.*;
```
oswaldquek, line 8: What's the reason for adding this line? Also we are for explicit importing for classes and avoid importing with `.*`.

oswaldquek (changes requested):
This PR looks like setting up ledger to run a webhook pact which is not published yet. I assume that's coming soon?

src/test/java/uk/gov/pay/ledger/pact/ConsumerContractTestSuite.java
nlsteers: We import every class in this package, hence *

nlsteers:
Ledger needs to know about Webhooks first before I can publish the pacts, `@IgnoreNoPactsToVerify` will prevent this failing in the meantime.

src/test/java/uk/gov/pay/ledger/pact/ConsumerContractTestSuite.java
oswaldquek: That may be true but it doesn't address what our coding standards/style might be. I suggest asking the wider dev team for input here.

nlsteers: I have replaced the wild card with individual imports.

oswaldquek (approved):

PR #2083, opened 2022-08-08
PP-9593: Change TotalAmount in report CSV

alan-gds:
This column has been causing confusion for some services, so we are changing how it is calculated:

* It remains the same as before for successful payments (amount, plus surcharge if present).
* It is now zero for all non-successful completed payments.
* It is now left blank for any payments still in progress.

These changes will also affect the TotalAmount column in the Payments To Your Bank Account download.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
@@ -99,7 +99,15 @@ public class CsvTransactionFactory {
 
                 result.put(FIELD_GOVUK_PAYMENT_ID, transactionEntity.getExternalId());
                 result.put(FIELD_AMOUNT, penceToCurrency(transactionEntity.getAmount()));
-                result.put(FIELD_TOTAL_AMOUNT, penceToCurrency(totalOrAmount));
+
+                if (transactionEntity.getState() == TransactionState.SUCCESS) {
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
```diff
@@ -99,7 +99,19 @@ public Map<String, Object> toMap(TransactionEntity transactionEntity) {
 
                 result.put(FIELD_GOVUK_PAYMENT_ID, transactionEntity.getExternalId());
                 result.put(FIELD_AMOUNT, penceToCurrency(transactionEntity.getAmount()));
-                result.put(FIELD_TOTAL_AMOUNT, penceToCurrency(totalOrAmount));
+
+                if (transactionEntity.getState() == TransactionState.SUCCESS) {
+                    result.put(FIELD_TOTAL_AMOUNT, penceToCurrency(totalOrAmount));
+                }
+                else {
```
stephencdaly, line 106: We can combine this to just be `else if` to avoid nesting? 

Also, we put `else` on the same line as the closing `}`. We follow whatever rules intelliJ uses to format code, so best to just press `Command + Option + L` before you commit a file.

stephencdaly (commented):
Looks good! Very minor style comment

stephencdaly (approved):

PR #2204, opened 2022-10-05
PP-10182: Endpoint to redact reference

oswaldquek:


src/main/java/uk/gov/pay/ledger/transaction/service/TransactionService.java
```diff
@@ -39,8 +37,9 @@
 
 public class TransactionService {
 
+    public static final String REDACTED_REFERENCE_NUMBER = "****";
```
sfount, line 40: As this isn't actually used by the code I think it's probably best to
* remove this and update the tests to just check for the `****` string (my vote)
* use this constant in the update sql 

Probably best to not have something in there that's only used by tests

sfount (commented):
Looks great, scoping the update to the specific JSONB field feels much clearer/ cleaner. One small comment

sfount (approved):

PR #2340, opened 2022-12-16
PP-10437 Add user identifier column to agreements table

sfount:
The user identifier is currently included in the event stream when agreements are created, this data needs to be represented on the Ledger agreement resource and will then be used by the API and the admin tool.

Add a column to the agreements projection table to allow the object mapper to appropriately set the `user_identifier` attribute.

Index the user identifier using a standard btree index, this will allow us filter agreements by this value if required in the future. It's likely that most agreements won't come with user identifiers initially so this shouldn't use up much space.


james-peacock-gds (approved):
LGTM!

PR #2341, opened 2022-12-16
PP-10437 Serialisation for user identifier on agreements 

sfount:
Depends on https://github.com/alphagov/pay-ledger/pull/2340

Add `userIdentifier` to the agreement entity which maps to the new
`user_identifier` column introduced in
https://github.com/alphagov/pay-ledger/pull/2340.

Set mappings between the agreement entity and agreement model, assert
that this is appropriately then shown from the agreement resource.


SandorArpa (approved):
LOKTM

PR #2383, opened 2023-01-10
PP-10485 Add test assertions that links are included

stephencdaly:
Add assertions to an integration tests for the search agreements endpoint to check that the response includes pagination links.


SandorArpa (approved):
LOKTM

PR #2384, opened 2023-01-10
PP-10445: Factor out some resource logic into a service

alan-gds:
Move the bulk of the logic for endpoint `/v1/agreement/{agreementExternalId}` into `AgreementService`, to facilitate testing.  We're about to make this more complex by adding some filters, so taking the existing logic out of the resource class is worthwhile.


sfount (dismissed):
Really good refactor! Code + tests are both a lot easier to read

sfount (approved):

PR #2386, opened 2023-01-10
PP-10263 Fix flakey dispute ITests

kbottla:
## WHAT
- For the dispute consumer pact tests, dispute events are sent to the SQS queue with `parent_resource_external_id`. No transaction exists for the `parent_resource_external_id`, so the dispute event projection fails and the message stays in the SQS queue.
- SQS queue message can then be processed later by another test which results in multiple dispute events in the database. One from the previous test and another for the current one. This causes tests to fail as the event returned by the database could be a different one.
- Dispute Itests also don't assert the transaction. At the time of writing the tests, we were not projecting dispute events.

## Changes
- Updated test to insert parent transaction at the time of setup
- Asserting transaction data too as we have projection enabled for dispute events


sfount (approved):
Good catch

PR #2390, opened 2023-01-11
PP-10445: Add filters to `/v1/agreement/{agreementExternalId}` endpoint

alan-gds:
* New filters are `account_id` and `service_id`
* Either or both of these must be provided, unless the `override_account_or_service_id_restriction` query param is submitted and set to true
* I've reordered some of the existing integration tests


sfount (approved):
Really good coverage on the headers/ query params that all combine to slightly different paths, looks great 👍

PR #2425, opened 2023-01-30
PP-10639: Map event types to `AgreementStates`

alan-gds:
We are going to remove the `status` field from all `AgreementEvent` implementations, so instead, map `SalientEventTypes` to `AgreementStates`.


src/main/java/uk/gov/pay/ledger/agreement/state/AgreementState.java
```diff
@@ -0,0 +1,67 @@
...
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum AgreementState {
+
+    CREATED("created"),
+    ACTIVE("active"),
+    CANCELLED("cancelled"),
+    EXPIRED("expired");
+
+    private final String oldStatus;
```
sfount, line 18: If we're not using old status, code or message I think it's OK to leave them out from this -- I think some of the old status pieces for transactions were only used to transition from a previous method of showing statuses to the current system and shouldn't be needed by this

sfount (commented):
Looks really good, I think a few bits could be removed from the state mapper and then this should be good to merge and map when agreements are created in the agreement factory

alan-gds:
I think AGREEMENT_SETUP (noun) is misnamed and should be AGREEMENT_SET_UP (verb phrase), but it needs to be kept in sync with what's in Connector, so for the moment I'll go with AGREEMENT_SETUP.  I'd like to rename it in a subsequent PR though.

src/main/java/uk/gov/pay/ledger/agreement/state/AgreementState.java
alan-gds: 👍

src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementsFactory.java
```diff
@@ -32,4 +35,10 @@ public PaymentInstrumentEntity createPaymentInstrument(EventDigest eventDigest)
         entity.setEventCount(eventDigest.getEventCount());
         return entity;
     }
+
+    private AgreementStatus getStatus(EventDigest eventDigest) {
```
stephencdaly, line 39: I think `AgreementsFactoryTest` can be updated to remove the `status` from the event payload, as it will no longer be used to resolve the status.

src/main/java/uk/gov/pay/ledger/agreement/state/AgreementState.java
```diff
@@ -0,0 +1,42 @@
...
+import com.fasterxml.jackson.annotation.JsonFormat;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import uk.gov.pay.ledger.event.model.SalientEventType;
+import uk.gov.service.payments.commons.model.agreement.AgreementStatus;
+
+import java.util.Map;
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonFormat(shape = JsonFormat.Shape.OBJECT)
+public enum AgreementState {
```
stephencdaly, line 12: Do we need both `AgreeementState` and `AgreementStatus` enums as it's a 1:1 mapping? I'd say that resolving the status should go directly in `AgreementStatus` unless there's a reason it wouldn't work.

stephencdaly (changes requested):

src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementsFactory.java
stephencdaly: We can also remove the "status" from `QueueMessageReceiverIT` tests for agreement event tests

src/main/java/uk/gov/pay/ledger/agreement/state/AgreementState.java
alan-gds: I wondered about this.  I'm happy to kill off `AgreementState` as it's currently redundant.  If we ever end up with anything else that needs to go in an `AgreementState`, we can always resurrect it.

src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementsFactory.java
```diff
@@ -32,4 +37,18 @@ public PaymentInstrumentEntity createPaymentInstrument(EventDigest eventDigest)
...
+        var eventType = eventDigest.getMostRecentSalientEventType().get();
+
+        return EVENT_TYPE_AGREEMENT_STATE_MAP.get(eventType);
+    }
+
+    private static final Map<SalientEventType, AgreementStatus> EVENT_TYPE_AGREEMENT_STATE_MAP =
```
stephencdaly, line 47: I think it would be better for this static map to live in the `AgreementStatus` enum so that it mirrors `TransactionState` and `PayoutState` to keep everything a bit more uniform.

stephencdaly (commented):
Looks good, just a very minor comment

src/main/java/uk/gov/pay/ledger/event/model/SalientEventType.java
```diff
@@ -54,7 +54,12 @@ public enum SalientEventType {
 
     PAYOUT_CREATED,
     PAYOUT_PAID,
-    PAYOUT_FAILED;
+    PAYOUT_FAILED,
+
+    AGREEMENT_CREATED,
+    AGREEMENT_SETUP,
```
stephencdaly, line 60: Should this now be `AGREEMENT_SET_UP`?

alan-gds: Yes!!!  I'll push a fix.

src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementsFactory.java
alan-gds: `AgreementStatus` comes from `pay-java-commons`, which doesn't know about `SalientEventType`.  I'm open to suggestions for where else the map could live though.

stephencdaly: Aha, I didn't spot that! In that case I think it is fine for this to live here. I guess the only minor change I might make would be to move this towards the top of the class where we usually define constants.

src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementsFactory.java
```diff
@@ -32,4 +45,10 @@ public PaymentInstrumentEntity createPaymentInstrument(EventDigest eventDigest)
         entity.setEventCount(eventDigest.getEventCount());
         return entity;
     }
+
+    private AgreementStatus getStatus(EventDigest eventDigest) {
+        var eventType = eventDigest.getMostRecentSalientEventType().get();
+
+        return EVENT_TYPE_AGREEMENT_STATE_MAP.get(eventType);
```
stephencdaly, line 52: We should never just call `get` on an Optional without checking whether the value is present, we should either return a default value or throw an exception. In this case, I think an exception would make sense as we would always expect one to be returned

```suggestion
        return eventDigest.getMostRecentSalientEventType().map(eventType -> EVENT_TYPE_AGREEMENT_STATE_MAP.get(eventType)).orElseThrow(() -> new RuntimeException("No salient event found for event digest for agreement with external ID: [xxxx]"));
```

src/main/java/uk/gov/pay/ledger/agreement/entity/AgreementsFactory.java
```diff
@@ -32,4 +45,10 @@ public PaymentInstrumentEntity createPaymentInstrument(EventDigest eventDigest)
         entity.setEventCount(eventDigest.getEventCount());
         return entity;
     }
+
+    private AgreementStatus getStatus(EventDigest eventDigest) {
+        return eventDigest.getMostRecentSalientEventType()
+                .map(EVENT_TYPE_AGREEMENT_STATE_MAP::get)
+                .orElseThrow(() -> new RuntimeException("No salient event found for event digest for agreement with external ID: [xxxx]"));
```
stephencdaly, line 52: Need to template in the agreement external ID - my suggestion was just an example of the exception message

alan-gds: D'oh, sorry

stephencdaly (approved):
LGTM!

PR #2438, opened 2023-02-03
PP-10488 Add agreement_external_id to MDC

stephencdaly:
- If the request has the `agreementExternalId` path paremeter, add it to the MDC to be included as a field for log lines.


james-peacock-gds (approved):
looks fine

PR #2503, opened 2023-03-08
PP-10701: Add `/v1/agreement/{agreementExternalId}/event` endpoint

alan-gds:
* Gets all events for an agreement and all of its associated payment instruments
* Includes events for old payment instruments that have been superseded by a new one
* A service Id must be provided in the request

The new `EventDto` class introduced in this PR is not named consistently with other DTO classes in this project.  Following this work, we will rename the existing `Event` class to `EventEntity`, and rename `EventDto` to `event`.  We will also remove the unused `/v1/event/{eventId}` endpoint, which is the only place that currently exposes the existing `Event` class.


src/main/java/uk/gov/pay/ledger/agreement/dao/AgreementDao.java
```diff
@@ -130,6 +148,18 @@ public Long getTotalForSearch(AgreementSearchParams searchParams) {
         });
     }
 
+    // Includes events for all associated payment instruments, including old payment instruments that have been replaced.
+    public List<Event> findAssociatedEvents(AgreementEntity agreementEntity, String serviceId) {
```
kbottla, line 152: Can we just pass agreement external ID as entity entity object is not needed

src/main/java/uk/gov/pay/ledger/agreement/dao/AgreementDao.java
```diff
@@ -87,6 +89,22 @@ public class AgreementDao {
             "FROM agreement a " +
             ":searchExtraFields";
 
+    private static final String SELECT_ASSOCIATED_EVENTS = "SELECT results.* FROM " +
+            "(" +
+                "SELECT e.id, e.sqs_message_id, e.service_id, e.live, 'payment_instrument' AS resource_type_name, " +
+                "e.resource_external_id, e.parent_resource_external_id, e.event_date, e.event_type, e.event_data " +
+                "FROM agreement a " +
```
kbottla, line 96: We don't need to join agreement table as no data is needed from it. I think querying just by agreement ID should be fine

Join between payment_instrument and event table is an inner join. Worth checking if index on event table is being used for this query

src/main/java/uk/gov/pay/ledger/event/model/EventDto.java
```diff
@@ -0,0 +1,98 @@
...
+                    event.getResourceType(),
+                    event.getEventType(),
+                    event.getEventDate(),
+                    event.getLive(),
+                    event.getServiceId(),
+                    objectMapper.readValue(event.getEventData(), new TypeReference<>() {
+                    })
+            );
+        }
+        catch (IOException e) {
```
kbottla, line 63: It would be good to have tests for this (if not exists already) as there are a few different paths here

src/main/java/uk/gov/pay/ledger/agreement/service/AgreementService.java
```diff
@@ -111,7 +122,16 @@ public AgreementSearchResponse searchAgreements(AgreementSearchParams searchPara
                 .withPaginationBuilder(paginationBuilder);
     }
 
+    public List<EventDto> findEvents(String agreementExternalId, String serviceId) {
+        var agreementEntity = findAgreementEntity(agreementExternalId, true, null, serviceId)
```
kbottla, line 126: It would be good to test 404 exception too

src/main/java/uk/gov/pay/ledger/agreement/dao/AgreementDao.java
alan-gds: Yes, there is an index on event.resource_external_id.

No data is selected from the agreement table, but it is needed in the join because it is used for filtering by serviceId.  I could get rid of it if it is the case that the serviceId column in the event table will always be populated for payment instrument events.  I don't *think* this is the case, but I may be wrong - do you know?

alan-gds: On reflection, I can get rid of the join if I just stop checking the serviceId in the query entirely - it's already checked when we look up the agreement entity in the service (to ensure the agreement exists).

src/main/java/uk/gov/pay/ledger/event/model/EventDto.java
alan-gds: I've moved the return up into the catch, to make the flow clearer.  Do we still need tests?

src/main/java/uk/gov/pay/ledger/agreement/service/AgreementService.java
alan-gds: Agreed; done

src/main/java/uk/gov/pay/ledger/agreement/dao/AgreementDao.java
alan-gds: Yes, that makes sense.  I am still going to look up the entity so that we can return a 404 if it's not found, but we don't need to pass the whole thing to the DAO.

src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementResource.java
```diff
@@ -91,4 +89,22 @@ public Agreement get(@Parameter(description = "The unique external id for the ag
...
-}
\ No newline at end of file
+
+    @Path("/{agreementExternalId}/event")
+    @GET
+    @Timed
+    @Operation(
+            summary = "Find agreement events by agreement ID",
+            responses = {
+                    @ApiResponse(responseCode = "200", description = "OK", content = @Content(schema = @Schema(implementation = Agreement.class))),
```
kbottla, line 99: schema is `Agreement.class` but schema returned is `List<EventDto>`

src/main/java/uk/gov/pay/ledger/agreement/dao/AgreementDao.java
```diff
@@ -87,6 +89,21 @@ public class AgreementDao {
...
+                "FROM payment_instrument pi " +
+                "JOIN event e ON e.resource_external_id = pi.external_id " +
+                "WHERE pi.agreement_external_id = :externalId " +
+                "UNION SELECT " +
+                "e.id, e.sqs_message_id, e.service_id, e.live, 'agreement' AS resource_type_name, e.resource_external_id, " +
+                "e.parent_resource_external_id, e.event_date, e.event_type, e.event_data " +
+                "FROM event e " +
+                "WHERE e.resource_external_id = :externalId " +
+            ") AS results " +
+            "ORDER BY results.event_date";
```
kbottla, line 105: Can the order (asc, desc) be specified explicitly so it is clear when reading the query?

src/main/java/uk/gov/pay/ledger/agreement/resource/AgreementResource.java
alan-gds: Oops, good spot.  I've changed it to `AgreementEventsResponse.class`, by analogy with a similar endpoint in `TransactionResource`.

src/main/java/uk/gov/pay/ledger/agreement/dao/AgreementDao.java
alan-gds: Happy to make this more explicit - done.

PR #2632, opened 2023-05-10
PP-10762: Allow ledger to use an overriden SNS endpoint for local testing

jfharden:
Allow the SNS client to be built with an overridden endpoint to allow local testing.

Note that I have matched the naming used in the very similar SQS config, and have also matched the indentation style to that even though it feels very strange to me (but then I'm not a java person 😄)

One big difference is SNS is using (and is already using) aws-sdk v2, and SQS is using aws-sdk v1, so the actual client construction and configuration is quite different.


oswaldquek (approved):

jfharden:
I've got pay local setup so that ledger will publish to SNS and have used that to confirm this override works as intended.

PR #2657, opened 2023-05-23
PP-10982 add can_retry to external transaction state

SandorArpa:
- add canRetry to `ExternalTransactionState`
- implement including canRetry when getting/searching transactions (if applicable)
- add tests

### DIFF
--- a/openapi/ledger_spec.yaml
+++ b/openapi/ledger_spec.yaml
@@ -1281,6 +1281,12 @@ components:
     ExternalTransactionState:
       type: object
       properties:
+        can_retry:
+          type: boolean
+          description: Only applicable for failed recurring payments. Default value
+            is null
+          example: true
+          nullable: true
         code:
           type: string
           example: P0030
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/model/TransactionView.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/model/TransactionView.java
@@ -184,8 +184,7 @@ public class TransactionView {
                     .withAgreementId(payment.getAgreementId())
                     .withDisputed(payment.getDisputed());
             if (payment.getState() != null) {
-                paymentBuilder = paymentBuilder
-                        .withState(ExternalTransactionState.from(payment.getState(), statusVersion));
+                paymentBuilder = paymentBuilder.withState(ExternalTransactionState.from(payment.getState(), statusVersion, payment.getCanRetry()));
             }
             return paymentBuilder.build();
         }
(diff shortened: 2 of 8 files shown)


openapi/ledger_spec.yaml
```diff
@@ -1281,6 +1281,10 @@ components:
     ExternalTransactionState:
       type: object
       properties:
+        can_retry:
+          type: boolean
+          default: false
```
kbottla, line 1286: as per https://github.com/alphagov/pay-ledger/pull/2657/files#diff-9c8dd90f4e46450f6d1b524141213d176f435827db34b3705b60e704fff438caR58 default is null but false here

src/main/java/uk/gov/pay/ledger/transaction/search/model/TransactionView.java
```diff
@@ -184,8 +184,11 @@ public static TransactionView from(Transaction transaction, int statusVersion) {
                     .withAgreementId(payment.getAgreementId())
                     .withDisputed(payment.getDisputed());
             if (payment.getState() != null) {
-                paymentBuilder = paymentBuilder
-                        .withState(ExternalTransactionState.from(payment.getState(), statusVersion));
+                if (payment.getAuthorisationMode() == AuthorisationMode.AGREEMENT && payment.getCanRetry() != null) {
```
kbottla, line 187: Is it possible to use one static method without having to check for authorisation_mode ?

`ExternalTransactionState.from(payment.getState(), statusVersion, payment.getCanRetry())`

for one-off can_retry field will always be null and will be excluded

src/main/java/uk/gov/pay/ledger/transaction/state/ExternalTransactionState.java
```diff
@@ -47,13 +55,19 @@ public String getMessage() {
...
+    }
+
     @Override
     public String toString() {
         return "ExternalTransactionState{" +
                 "value='" + value + '\'' +
                 ", finished=" + finished +
                 ", code='" + code + '\'' +
                 ", message='" + message + '\'' +
+                ", can_retry=" + canRetry +
```
kbottla, line 70: This will be null most of the time. Can this be added only when it is non-null?

openapi/ledger_spec.yaml
SandorArpa: 👍

src/main/java/uk/gov/pay/ledger/transaction/search/model/TransactionView.java
SandorArpa: Yes, although we do check everywhere for authorisation mode before using canRetry

src/main/java/uk/gov/pay/ledger/transaction/state/ExternalTransactionState.java
SandorArpa: 👍

kbottla (approved):

PR #2673, opened 2023-05-31
PP-10509 Don't publish agreement events to SNS

kbottla:
## WHAT 
- Currently, Ledger is attempting to publish events other than disputes to `card_payment_events` SNS topic. We can restrict this to PAYMENT/REFUNDS events as consumers (webhooks/adminusers) of SNS do not handle events other than PAYMENT/REFUND/DISPUTE.
- Also Ledger currently fails to publish agreement events to SNS and logs `Failed to publish event for message` due to `NullPointerException` when getting `rawMessageBody` (as it is only available for events from SQS but not agreements which are sent synchronously)


SandorArpa (approved):
LOKTM

PR #2677, opened 2023-06-01
PP-10509 Get EventDigest by resource ID and type

kbottla:
## WHAT
- Currently, when projecting agreement (during GET), events for the resource_external_id are processed and mapped to the agreement.
- If events for the resource_external_id are not of type AGREEMENT, Ledger still considers them and attempts to project events which cause an exception (`java.lang.RuntimeException: No salient event found for event digest for agreement with external ID`). This is because agreement status cannot be derived from these events
- So filtering events by resource type AGREEMENT as well. If no events exist, it will result in 404 which is expected
- Refactored `EventMessageHandler.processSingleMessage` method to extract publishing to SNS functionality as the method is getting complex.


james-peacock-gds (approved):
LGTM

PR #2721, opened 2023-06-20
PP-11121 Add PACT state for an agreement

james-peacock-ho:
- Add PACT provider state, an agreement with payment instrument exists


SandorArpa (approved):
LOKTM

PR #2726, opened 2023-06-21
PP-11121 fix pact test

SandorArpa:
- add event count to agreement
- set agreement as not live


james-peacock-gds (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/) Pay

PR #2736, opened 2023-06-26
PP-9884 refactor agreement pact tests

SandorArpa:
- refactor 3 agreements so selfservice can use it
- refactor agreement with payment instrument so we can param reference

 with @james-peacock-gds


james-peacock-gds (approved):
I'm confident that this change does not require further review and meets the secure coding requirements of [GOV.UK](http://gov.uk/) Pay

PR #2766, opened 2023-07-14
PP-11152 Update pact state

stephencdaly:
Update Pact state that sets up recurring payments to set up a transaction with a particular external ID, extracted from the params specified by the Pact test.

This is so that we can add a Pact test to connector to get a single recurring payment transaction.


SandorArpa (approved):
LGTM

PR #2777, opened 2023-07-21
PP-11246 Set mandatory fields for events

james-peacock-gds:
- Make timestamp, externalId, resourceType and eventType mandatory for post events endpoint
- update openapi spec

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
+++ b/src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
@@ -11,6 +11,7 @@ import uk.gov.pay.ledger.event.model.ResourceType;
 import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeDeserializer;
 import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeSerializer;
 
+import javax.validation.constraints.NotNull;
 import java.time.ZonedDateTime;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
@@ -21,19 +22,24 @@ public class EventMessageDto {
 
     private Boolean live;
 
+    @NotNull
     @JsonDeserialize(using = MicrosecondPrecisionDateTimeDeserializer.class)
     private ZonedDateTime timestamp;
 
+    @NotNull
     @JsonProperty("resource_external_id")
     private String externalId;
 
     @JsonProperty("parent_resource_external_id")
     private String parentExternalId;
 
+    @NotNull
     public String eventType;
 
+    @NotNull
     public ResourceType resourceType;
 
+    @NotNull
     @JsonProperty("event_details")
     private JsonNode eventData;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
```diff
@@ -21,19 +22,24 @@ public class EventMessageDto {
 
     private Boolean live;
 
+    @NotNull
```
kbottla, line 25: Can we add test(s) for these changes?

james-peacock-gds: have added an integration test to demonstrate that a request missing one of the mandatory fields will be rejected with 422. Is this sufficient?

src/test/java/uk/gov/pay/ledger/event/resource/EventResourceAgreementIT.java
```diff
@@ -85,4 +76,45 @@ public void shouldWriteCancelledEvent() {
...
+                        .put("cancelled_date", now.toString()))
+                .put("resource_external_id", agreementId);
+        params.put(event);
+
+        given().port(port)
+                .contentType(JSON)
+                .body(params.toString())
+                .post("/v1/event")
+                .then()
+                .statusCode(422);
```
kbottla, line 104: I think testing 422 here is fine but not clear which field is supposed to be mandatory as per the test name. 

Would be good to update the test name or add a few more assertions for the expected fields without relying on the comment.

src/test/java/uk/gov/pay/ledger/event/resource/EventResourceAgreementIT.java
```diff
@@ -85,4 +76,45 @@ public void shouldWriteCancelledEvent() {
         assertThat(agreementEntity.getCancelledByUserEmail(), is("jdoe@example.org"));
         assertThat(agreementEntity.getCancelledDate(), is(ZonedDateTimeMatchers.within(1, ChronoUnit.SECONDS, now)));
     }
+
+    @Test
+    public void shouldNotWriteEventIfAMandatoryFieldIsNotPresent() {
```
kbottla, line 81: This doesn't cover fields other than event_type. We could add a few unit tests for other fields. 

A good place to add unit tests is https://github.com/alphagov/pay-ledger/blob/master/src/test/java/uk/gov/pay/ledger/event/resource/EventResourceTest.java#L24

kbottla, line 81: Just noticed that this is in `EventResourceAgreementIT`. This test class is specifically testing for agreement projection. We can add a test to `EventResourceIT`.

And looks like there is already a test which could be improved https://github.com/alphagov/pay-ledger/blob/10b3c7dfb45475395bd6745dbd514dc8e47008a6/src/test/java/uk/gov/pay/ledger/event/resource/EventResourceIT.java#L62 as doesn't verify status or error message being returned.

PR #2785, opened 2023-07-25
PP-11246 Set mandatory fields for events

james-peacock-gds:
- Make timestamp, externalId, resourceType and eventType mandatory for post events endpoint
- Update openapi spec
- Add unit tests
- Add response status code check to integration test

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
+++ b/src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
@@ -11,6 +11,7 @@ import uk.gov.pay.ledger.event.model.ResourceType;
 import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeDeserializer;
 import uk.gov.service.payments.commons.api.json.MicrosecondPrecisionDateTimeSerializer;
 
+import javax.validation.constraints.NotNull;
 import java.time.ZonedDateTime;
 
 @JsonIgnoreProperties(ignoreUnknown = true)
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
```diff
@@ -21,19 +22,24 @@ public class EventMessageDto {
 
     private Boolean live;
 
+    @NotNull
     @JsonDeserialize(using = MicrosecondPrecisionDateTimeDeserializer.class)
     private ZonedDateTime timestamp;
 
+    @NotNull
```
kbottla, line 29: We can add text to annotations so the error message is clear. With `@NonNull` annotations, the error message returned will be generic so it will be unclear what must not be null.

```
{
    "error_identifier": "GENERIC",
    "message": [
        "must not be null"
    ]
}
```

Can add something like `@NotNull(message = "Field [resource_external_id] cannot be null")` so the error message will be clear as below and in the test, we can assert for the presence of specific text `Field [resource_external_id] cannot be null` in the response

```
{
    "error_identifier": "GENERIC",
    "message": [
        "Field [resource_external_id] cannot be null"
    ]
}
```

For multiple fields, it will look like

```
{
    "error_identifier": "GENERIC",
    "message": [
        "Field [timestamp] cannot be null",
        "Field [resource_external_id] cannot be null"
    ]
}
```

james-peacock-gds: thank you!

kbottla (approved):

src/main/java/uk/gov/pay/ledger/queue/EventMessageDto.java
```diff
@@ -21,19 +22,24 @@ public class EventMessageDto {
...
     @JsonProperty("parent_resource_external_id")
     private String parentExternalId;
 
+    @NotNull(message = "Field [event_type] cannot be null")
     public String eventType;
 
+    @NotNull(message = "Field [resource_type] cannot be null")
     public ResourceType resourceType;
 
+    @NotNull(message = "Field [event_details] cannot be null")
```
kbottla, line 42: `event_details` can be empty. Some events don't carry any payload. They just update the state of the resource.

kbottla (approved):

kbottla (approved):

PR #2884, opened 2023-09-19
PP-11548: Initialise prom metrics without default sample builder

jfharden:
# What?
Now we don't need to build our default labels in the app (since they are now added by the sidecar) we should remove the default sample builder.

1. Remove default sample builder
2. Update local build script to account for m1 macs, and use the correct container tag

# How?

This change has already been rolled out to connector and is working well.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/app/LedgerApp.java
+++ b/src/main/java/uk/gov/pay/ledger/app/LedgerApp.java
@@ -31,7 +31,6 @@ import uk.gov.pay.ledger.queue.managed.QueueMessageReceiver;
 import uk.gov.pay.ledger.report.resource.PerformanceReportResource;
 import uk.gov.pay.ledger.report.resource.ReportResource;
 import uk.gov.pay.ledger.transaction.resource.TransactionResource;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/ledger/app/LedgerApp.java
```diff
@@ -71,8 +70,11 @@ public DataSourceFactory getDataSourceFactory(LedgerConfig configuration) {
 
     @Override
     public void run(LedgerConfig config, Environment environment) {
-        config.getEcsContainerMetadataUriV4().ifPresent(uri -> initialisePrometheusMetrics(environment, uri));
-        
+        LOGGER.info("Initialising prometheus metrics.");
```
nlsteers, line 73: how useful is this log line?

nlsteers (dismissed):
Looks good, can probably remove the logging call though.

src/main/java/uk/gov/pay/ledger/app/LedgerApp.java
jfharden: Not, good call, removed

jfharden:
@nlsteers I've removed the log line, would you mind reapproving

nlsteers (approved):

PR #2894, opened 2023-09-21
PP-11232 Add method to redact PII from transaction

kbottla:
## WHAT
- Added a new DAO method to redact PII (personally identifiable information) from transactions
- Uses JSONB_SET with `false` as the last arg (`create_missing` - default is true) to replace PII from the transaction_details JSON blob.
    - can't replace multiple fields at once so using nested JSONB_SET for multiple columns.
- Removed fields `reference, cardholder_name, email` from transaction_details JSON as these are not used anywhere and are also top-level DB columns. So having them in transaction_details as `<REDACTED>` doesn't add any value

- Note: `<REDACTED>` string is not finalised and may be updated later. So some other fields first/last digits card number

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
@@ -213,6 +213,18 @@ public class TransactionDao {
                     "agreement_id = EXCLUDED.agreement_id " +
                     "WHERE EXCLUDED.event_count >= transaction.event_count;";
 
+    private static final String REDACT_PII_FROM_TRANSACTIONS =
+            "UPDATE transaction t " +
+                    "set reference = '<REDACTED>'," +
+                    "  cardholder_name = (case when cardholder_name IS NOT NULL then '<REDACTED>' end), " +
+                    "  email = (case when email IS NOT NULL then '<REDACTED>' end), " +
+                    "  transaction_details = JSONB_SET(" +
+                    "                              JSONB_SET(transaction_details, '{address_line1}','\"<REDACTED>\"', false), " +
+                    "                              '{address_line2}','\"<REDACTED>\"', false" +
+                    "                        )" +
+                    "                         -'{reference,cardholder_name,email}'::text[]" +
+                    " WHERE t.external_id = :externalId";
+
     private static final String GET_SOURCE_TYPE_ENUM_VALUES =
             "SELECT " +
                     "pg_enum.enumlabel " +
@@ -375,6 +387,13 @@ public class TransactionDao {
                     .map(new TransactionMapper())
                     .list();
         });
+    }
+
+    public void redactPIIFromTransaction(String transactionExternalId) {
+        jdbi.withHandle(handle ->
+                handle.createUpdate(REDACT_PII_FROM_TRANSACTIONS)
+                        .bind("externalId", transactionExternalId)
+                        .execute());
     }
 
     private String createSearchTemplate(TransactionSearchParams searchParams, String baseQueryString) {
--- a/src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
@@ -1,6 +1,11 @@
 package uk.gov.pay.ledger.transaction.dao;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
+import com.google.gson.JsonObject;
+import org.hamcrest.Matchers;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
@@ -14,7 +19,6 @@ import uk.gov.pay.ledger.util.fixture.TransactionFixture;
 import uk.gov.service.payments.commons.model.Source;
 
 import java.time.ZonedDateTime;
-import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.List;
 import java.util.stream.Collectors;
@@ -41,6 +45,7 @@ class TransactionDaoIT {
     public static AppWithPostgresAndSqsExtension rule = new AppWithPostgresAndSqsExtension();
 
     private TransactionDao transactionDao = new TransactionDao(rule.getJdbi(), mock(LedgerConfig.class));
+    private ObjectMapper objectMapper = new ObjectMapper();
 
     @Test
     void shouldInsertTransaction() {
@@ -455,8 +460,8 @@ class TransactionDaoIT {
     }
 
     @Nested
-    @DisplayName("TestFindTransactionsForRedaction")
-    class TestFindTransactionsForRedaction {
+    @DisplayName("FindTransactionsForRedaction")
+    class TestTransactionsForRedaction {
 
         @Test
         void shouldReturnTransactionsCorrectlyForDateRangesAndNumberOfTransactions() {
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -213,6 +213,18 @@ public class TransactionDao {
...
+    private static final String REDACT_PII_FROM_TRANSACTIONS =
+            "UPDATE transaction t " +
+                    "set reference = '<REDACTED>'," +
+                    "  cardholder_name = (case when cardholder_name IS NOT NULL then '<REDACTED>' end), " +
+                    "  email = (case when email IS NOT NULL then '<REDACTED>' end), " +
+                    "  transaction_details = JSONB_SET(" +
+                    "                              JSONB_SET(transaction_details, '{address_line1}','\"<REDACTED>\"', false), " +
+                    "                              '{address_line2}','\"<REDACTED>\"', false" +
+                    "                        )" +
+                    "                         -'{reference,cardholder_name,email}'::text[]" +
```
oswaldquek, line 225: Because I'm not familiar with this syntax, what is the purpose of `-'{reference,cardholder_name,email}'::text[]`?

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
```diff
@@ -213,6 +213,18 @@ public class TransactionDao {
                     "agreement_id = EXCLUDED.agreement_id " +
                     "WHERE EXCLUDED.event_count >= transaction.event_count;";
 
+    private static final String REDACT_PII_FROM_TRANSACTIONS =
+            "UPDATE transaction t " +
+                    "set reference = '<REDACTED>'," +
+                    "  cardholder_name = (case when cardholder_name IS NOT NULL then '<REDACTED>' end), " +
+                    "  email = (case when email IS NOT NULL then '<REDACTED>' end), " +
```
oswaldquek, line 220: What happens to the update if the cardholder_name and/or email is null?

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -505,4 +510,91 @@ void shouldReturnTransactionsCorrectlyForDateRangesAndNumberOfTransactions() {
...
+                    .withCardholderName("Joe B")
+                    .withTransactionDetails(transactionDetails1.toString())
+                    .insert(rule.getJdbi())
+                    .toEntity();
+
+            JsonObject transactionDetails2 = new JsonObject();
+            transactionDetails2.addProperty("reference", "ref-2");
+            transactionDetails2.addProperty("cardholder_name", "Jane D");
+            transactionDetails2.addProperty("email", "test2@email.com");
+            TransactionEntity transactionThatShouldNotBeRedacted = aTransactionFixture()
```
oswaldquek, line 540: What makes this a transaction that should not be redacted?

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -505,4 +510,91 @@ void shouldReturnTransactionsCorrectlyForDateRangesAndNumberOfTransactions() {
...
+            assertThat(jsonNode.get("reference"), is(Matchers.nullValue()));
+            assertThat(jsonNode.get("cardholder_name"), is(nullValue()));
+            assertThat(jsonNode.get("email"), is(nullValue()));
+            assertThat(jsonNode.get("address_line1").asText(), is("<REDACTED>"));
+            assertThat(jsonNode.get("address_line2").asText(), is("<REDACTED>"));
+
+            TransactionEntity transactionEntityNotRedacted = transactionDao.findTransactionByExternalId(transactionThatShouldNotBeRedacted.getExternalId()).get();
+            assertThat(transactionEntityNotRedacted.getReference(), is("ref-2"));
+            assertThat(transactionEntityNotRedacted.getCardholderName(), is("Jane D"));
+            assertThat(transactionEntityNotRedacted.getEmail(), is("test2@email.com"));
```
oswaldquek, line 567: It would help with readability if this were another test case

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
```diff
@@ -505,4 +510,91 @@ void shouldReturnTransactionsCorrectlyForDateRangesAndNumberOfTransactions() {
...
+
+            TransactionEntity transactionEntity = transactionDao.findTransactionByExternalId(transactionToRedact.getExternalId()).get();
+
+            assertThat(transactionEntity.getReference(), is("<REDACTED>"));
+            assertThat(transactionEntity.getCardholderName(), is(nullValue()));
+            assertThat(transactionEntity.getEmail(), is(nullValue()));
+
+            JsonNode jsonNode = objectMapper.readTree(transactionEntity.getTransactionDetails());
+
+            assertThat(jsonNode.get("reference"), is(nullValue()));
```
oswaldquek, line 593: Why is this reference null when it was added above?
```
            JsonObject transactionDetails1 = new JsonObject();
            transactionDetails1.addProperty("reference", "ref-1");
```

oswaldquek (changes requested):

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
kbottla: This will remove fields from `transaction_details` JSON data.

Mentioned in PR desc

Also mentioned in PR desc

> Removed fields `reference, cardholder_name, email` from transaction_details JSON as these are not used anywhere and are also top-level DB columns. So having them in transaction_details as `<REDACTED>` doesn't add any value

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
kbottla: They will stay null in that case. This test case should cover it https://github.com/alphagov/pay-ledger/pull/2894/files#diff-dcd3f9f3fa9bb7ad5ba913ef2f3e3b6c846e994cf7bdd9690cda980ed3410c29R571-R573

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
kbottla: This is because the field is being removed from transaction_details JSON 

as mentioned in PR desc

> Removed fields reference, cardholder_name, email from transaction_details JSON as these are not used anywhere and are also top-level DB columns. So having them in transaction_details as <REDACTED> doesn't add any value

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
kbottla: DAO method is to redact PII for a single transaction. This is another transaction that should not be selected for redaction.

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
kbottla: I am trying to assert that if another transaction is present it should not be redacted (that the = in SQL works for a single tx). Will add a new test for it

src/main/java/uk/gov/pay/ledger/transaction/dao/TransactionDao.java
oswaldquek: Ah, the `-` means "minus these fields".

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
oswaldquek: Ahhh ok.

src/test/java/uk/gov/pay/ledger/transaction/dao/TransactionDaoIT.java
oswaldquek: Hmm given your [explanation](https://github.com/alphagov/pay-ledger/pull/2894#discussion_r1332882033), I think it's fine to leave this test as is.

oswaldquek (approved):

PR #2895, opened 2023-09-21
PP-11232 Add DAO method to remove events

kbottla:
## WHAT
- Added DAO method to remove events for the list of transactions provided


jfharden (approved):
👍 Nice 👍

PR #2900, opened 2023-09-25
PP-11232 Add service method to expunge/redact historical data

kbottla:
## WHAT
- Added a service method to expunge/redact historical data
- Adds the core structure and logs when the flag is disabled. It also sends metrics for the total duration of the task.


jfharden (approved):
Looks good 👍 Loving the histogram :)

PR #2908, opened 2023-09-26
PP-11232 Update service to redact PII from transactions

kbottla:
## WHAT
- Redacts PII from transactions and deletes related events
- redact PII.. method processes transactions for a date range
    - start date: created_date of last processed transaction. If not gets the first transaction created date. When there are no transactions, uses the derived date up to which transactions can be redacted based on config. 
    - end date: Date up to which transactions can be redacted. This is based on config (default is now() - 7 years)
- Uses the system clock provided so fixed dates can be used for testing

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
+++ b/src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
@@ -6,16 +6,33 @@ import org.slf4j.LoggerFactory;
 import uk.gov.pay.ledger.app.LedgerConfig;
 import uk.gov.pay.ledger.app.config.ExpungeOrRedactHistoricalDataConfig;
 import uk.gov.pay.ledger.event.dao.EventDao;
+import uk.gov.pay.ledger.expungeorredact.dao.TransactionRedactionInfoDao;
 import uk.gov.pay.ledger.transaction.dao.TransactionDao;
+import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
 
 import javax.inject.Inject;
+import java.time.Clock;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import static java.lang.Math.min;
+import static java.time.ZoneOffset.UTC;
+import static java.util.Optional.ofNullable;
+import static net.logstash.logback.argument.StructuredArguments.kv;
+import static uk.gov.service.payments.logging.LoggingKeys.RESOURCE_EXTERNAL_ID;
 
 public class ExpungeOrRedactService {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ExpungeOrRedactService.class);
+    private final TransactionRedactionInfoDao transactionRedactionInfoDao;
     private final ExpungeOrRedactHistoricalDataConfig expungeOrRedactHistoricalDataConfig;
-    private TransactionDao transactionDao;
-    private EventDao eventDao;
+    private final Clock clock;
+    private final TransactionDao transactionDao;
+    private final EventDao eventDao;
+
+    private static final int PAGE_SIZE = 500;
 
     private static final Histogram duration = Histogram.build()
             .name("expunge_and_redact_historical_data_job_duration_seconds")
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactServiceTest.java
```diff
@@ -85,4 +104,88 @@ void shouldObserveJobDurationForMetrics() {
...
+    }
+
+    @Test
+    void shouldUseTheDateDerivedBasedOnConfigForTheStartingIndexWhenNoTransactionsExists() {
+        when(mockExpungeOrRedactHistoricalDataConfig.isExpungeAndRedactHistoricalDataEnabled()).thenReturn(true);
+        when(mockExpungeOrRedactHistoricalDataConfig.getNoOfTransactionsToRedact()).thenReturn(100);
+        when(mockExpungeOrRedactHistoricalDataConfig.getExpungeOrRedactDataOlderThanDays()).thenReturn(1);
+
+        when(mockTransactionRedactionInfoDao.getCreatedDateOfLastProcessedTransaction()).thenReturn(null);
+        when(mockTransactionDao.getCreatedDateOfFirstTransaction()).thenReturn(Optional.ofNullable(null));
```
oswaldquek, line 183: Is this a valid scenario? We would never not have a transaction at this point.

src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
```diff
@@ -6,38 +6,128 @@
...
     private static final Logger LOGGER = LoggerFactory.getLogger(ExpungeOrRedactService.class);
+    private final TransactionRedactionInfoDao transactionRedactionInfoDao;
     private final ExpungeOrRedactHistoricalDataConfig expungeOrRedactHistoricalDataConfig;
-    private TransactionDao transactionDao;
-    private EventDao eventDao;
+    private final Clock clock;
+    private final TransactionDao transactionDao;
+    private final EventDao eventDao;
+
+    private static final int PAGE_SIZE = 500;
```
oswaldquek, line 35: Why do we need this PAGE_SIZE? If we're concerned about the performance of having too many transactions to redact, wouldn't we adjust ExpungeOrRedactHistoricalDataConfig's noOfTransactionsToRedact accordingly?

src/test/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactServiceTest.java
kbottla: yeah not in test-prod environments but likely in a local dev environment

src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
kbottla: PAGE_SIZE is to limit the number of transactions we query from the database at a time and we most likely will get a constant performance time. This also limits the number of events deleted at a time which will be ~20 times no.of transactions.

If we only use `noOfTransactionsToRedact`, changing it directly impacts the performance of queries and we may not be able to redact as many transactions as required if there are any performance issues. `noOfTransactionsToRedact` will increase over time as more services join and with additional payment channels so the default value of 2500 could go high anytime.

oswaldquek: I suppose that's a valid reason. 

How often will `/v1/tasks/delete-or-redact-historical-data` be called though? If it's called frequently enough do we even need `noOfTransactionsToRedact`? I'm trying to achieve a simpler piece of code by avoiding a while loop, if possible. 

If we really need the while loop, could we have a test where the following is tested to ensure the while loop is tested properly?
```
verify(mockTransactionDao, times(2)).findTransactionsForRedaction(any(), any(), anyInt())
```

kbottla: Scheduled jobs can be run at half-an-hour/one-hour intervals. Job could overlap if they are run more frequently. while loop more flexibility to change job schedule and noOfTransactionsToRedact.

Added a test with different responses for `findTransactionsForRedaction`

oswaldquek: Hmm personally I think this might be bordering on over engineering here; we're making assumptions about jobs overlapping and performance tuning for which we haven't tested. I'm a big fan of simple code. I think it's worth getting a second opinion?

kbottla: Don't think it is over-engineering. We still need to paginate for better performance.

jfharden: I think I'm with krishna on this, we know from experience that our other jobs don't complete in test perf and eventually in staging, they would be much more likely to succeed if they paged (since they would successfully clean up some each run, instead of failing and leaving everything in the current state).

We also know in a few years time we'll be redacting large volumes of data and limiting the amount loaded into memory at once is very desirable

jfharden (approved):

src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
oswaldquek: Thanks for your input!

PR #2916, opened 2023-09-28
PP-11232 Move Itests out of nested test classes

kbottla:
## WHAT
- for unknown reasons, Github actions are not starting Postgres correctly for nested test classes and failing integration tests. Moving them out of nested test classes.


oswaldquek (approved):
A real shame we can't use nested test classes; let's hope this fixes the build.

PR #2923, opened 2023-09-29
PP-11232 Update search to find transactions to redact

kbottla:
## WHAT
- Updated search conditions to find transactions for redaction. Removed `=` when matching with `dateOfLastProcessedTransaction` so transaction processed in the last run is excluded in the new run

- Updates `transaction_redaction_info` after every batch of transactions processed. This is so that if the task is stopped midway (possible with perf env) it can resume after the last batch of txs processed.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
+++ b/src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
@@ -13,12 +13,12 @@ import uk.gov.pay.ledger.transaction.entity.TransactionEntity;
 import javax.inject.Inject;
 import java.time.Clock;
 import java.time.ZonedDateTime;
-import java.time.temporal.ChronoUnit;
 import java.util.List;
 import java.util.stream.Collectors;
 
 import static java.lang.Math.min;
 import static java.time.ZoneOffset.UTC;
+import static java.time.temporal.ChronoUnit.DAYS;
 import static java.util.Optional.ofNullable;
 import static net.logstash.logback.argument.StructuredArguments.kv;
 import static uk.gov.service.payments.logging.LoggingKeys.RESOURCE_EXTERNAL_ID;
@@ -110,6 +110,7 @@ public class ExpungeOrRedactService {
             createdDateOfLastProcessedTransaction = transactionsForRedaction
                     .stream().map(TransactionEntity::getCreatedDate)
                     .max(ZonedDateTime::compareTo).get();
+            transactionRedactionInfoDao.update(createdDateOfLastProcessedTransaction);
         }
 
         LOGGER.info("Completed redacting PII from transactions",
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
```diff
@@ -118,27 +119,24 @@ private void redactPIIFromTransactionsAndDeleteRelatedEvents() {
...
 
     private ZonedDateTime getCreatedDateOfLastProcessedTransaction() {
         return ofNullable(transactionRedactionInfoDao.getCreatedDateOfLastProcessedTransaction())
-                .orElseGet(this::getCreatedDateOfFirstTransaction)
+                .orElseGet(this::getDateBeforeCreatedDateOfFirstTransaction)
                 .withZoneSameInstant(UTC);
     }
 
-    private ZonedDateTime getCreatedDateOfFirstTransaction() {
+    private ZonedDateTime getDateBeforeCreatedDateOfFirstTransaction() {
```
oswaldquek, line 136: Should this be `getDayBeforeCreatedDateOfFirstTransaction` instead? Also why is this being changed to the day before the created date of the first transaction?

src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
```diff
@@ -118,27 +119,24 @@ private void redactPIIFromTransactionsAndDeleteRelatedEvents() {
...
         ZonedDateTime dateOfFirstTransaction = transactionDao.getCreatedDateOfFirstTransaction()
-                .orElseGet(this::getRedactTransactionsUpToDate);
+                .orElseGet(this::getRedactTransactionsUpToDate)
```
oswaldquek, line 138: I think calling `getRedactTransactionsUpToDate` in the `getDayBeforeCreatedDateOfFirstTransaction` method seems a bit confusing. Perhaps this call should be in the `getCreatedDateOfLastProcessedTransaction` method instead?

src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
kbottla: Updated name! 

DAO returns transaction **after** the last transaction processed date https://github.com/alphagov/pay-ledger/pull/2923/files#diff-257f1e610732a7f40031675387f1da0980f7c707f2704b901e8d87b1411bb6caR89 so for the first run this should be before the day so first transaction is included in the result.

src/main/java/uk/gov/pay/ledger/expungeorredact/service/ExpungeOrRedactService.java
kbottla: It is in the `getCreatedDateOfLastProcessedTransaction` as `transactionRedactionInfo` will need to be updated with the info created date of the first transaction. If this is not available (local dev env) it will be the date returned by `getRedactTransactionsUpToDate`. So for the first run date range will be  `getRedactTransactionsUpToDate()-1 to getRedactTransactionsUpToDate()`

oswaldquek (approved):

PR #2998, opened 2023-11-09
Pp 11612 illegal character in ledger url

hjvoid:
WHAT YOU DID?
The PaginationBuilder class was causing a 500 error if a user searched for a transaction with a curly braces character in the 'reference' field. At the point where ledgers `PaginationBuilder` builds the `.getSelfLink()` these characters were causing an `IllegalArgumentException` stemming from the `glassfish` package. I have altered the `uriWithParams` method to re-encode the parameters, with the exceptions of the `=` and `&` characters which are needed in their raw form for the response. 

As far as I can make out, glassfish is decoding the incoming api call somewhere within the code as it looks like SelfService is correctly sending the payload to ledger. However, so far I've been unsuccessful in finding exactly where in the code this happens.

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
@@ -20,8 +20,4 @@ public abstract class SearchParams {
     protected boolean isSet(CommaDelimitedSetParameter commaDelimitedSetParameter) {
         return commaDelimitedSetParameter != null && commaDelimitedSetParameter.isNotEmpty();
     }
-
-    public String encodeParam(String value) {
-        return URLEncoder.encode(value, StandardCharsets.UTF_8).replace("%3D", "=").replace("%26","&");
-    }
 }
--- a/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
@@ -2,9 +2,6 @@ package uk.gov.pay.ledger.common.search;
 
 import uk.gov.pay.ledger.util.CommaDelimitedSetParameter;
 
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
-
 public abstract class SearchParams {
 
     public abstract String buildQueryParamString(Long forPage);
--- a/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
@@ -2,6 +2,9 @@ package uk.gov.pay.ledger.common.search;
 
 import uk.gov.pay.ledger.util.CommaDelimitedSetParameter;
 
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
+
 public abstract class SearchParams {
 
     public abstract String buildQueryParamString(Long forPage);
@@ -16,5 +19,9 @@ public abstract class SearchParams {
 
     protected boolean isSet(CommaDelimitedSetParameter commaDelimitedSetParameter) {
         return commaDelimitedSetParameter != null && commaDelimitedSetParameter.isNotEmpty();
+    }
+
+    public String encodeParam(String value) {
+        return URLEncoder.encode(value, StandardCharsets.UTF_8).replace("%3D", "=").replace("%26","&");
     }
 }
--- a/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
@@ -9,6 +9,8 @@ import uk.gov.pay.ledger.util.CommaDelimitedSetParameter;
 
 import javax.ws.rs.DefaultValue;
 import javax.ws.rs.QueryParam;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -447,67 +449,65 @@ public class TransactionSearchParams extends SearchParams {
         List<String> queries = new ArrayList<>();
 
         if (accountIds != null && !accountIds.isEmpty()) {
-            queries.add(GATEWAY_ACCOUNT_EXTERNAL_FIELD + "=" + String.join(",", accountIds));
+            queries.add(GATEWAY_ACCOUNT_EXTERNAL_FIELD + "=" + URLEncoder.encode(String.join(",", accountIds), StandardCharsets.UTF_8));
         }
         if (isNotBlank(fromDate)) {
-            queries.add(FROM_DATE_FIELD + "=" + fromDate);
+            queries.add(FROM_DATE_FIELD + "=" + URLEncoder.encode(fromDate, StandardCharsets.UTF_8));
         }
         if (isNotBlank(toDate)) {
-            queries.add(TO_DATE_FIELD + "=" + toDate);
+            queries.add(TO_DATE_FIELD + "=" + URLEncoder.encode(toDate, StandardCharsets.UTF_8));
         }
-
         if (isNotBlank(email)) {
-            queries.add(EMAIL_FIELD + "=" + email);
+            queries.add(EMAIL_FIELD + "=" + URLEncoder.encode(email, StandardCharsets.UTF_8));
         }
         if (isNotBlank(reference)) {
-            queries.add(REFERENCE_FIELD + "=" + reference);
+            queries.add(REFERENCE_FIELD + "=" + URLEncoder.encode(reference, StandardCharsets.UTF_8));
         }
         if (isNotBlank(cardHolderName)) {
-            queries.add(CARDHOLDER_NAME_FIELD + "=" + cardHolderName);
+            queries.add(CARDHOLDER_NAME_FIELD + "=" + URLEncoder.encode(cardHolderName, StandardCharsets.UTF_8));
         }
         if (isNotBlank(firstDigitsCardNumber)) {
-            queries.add(FIRST_DIGITS_CARD_NUMBER_FIELD + "=" + firstDigitsCardNumber);
+            queries.add(FIRST_DIGITS_CARD_NUMBER_FIELD + "=" + URLEncoder.encode(firstDigitsCardNumber, StandardCharsets.UTF_8));
         }
         if (isNotBlank(lastDigitsCardNumber)) {
-            queries.add(LAST_DIGITS_CARD_NUMBER_FIELD + "=" + lastDigitsCardNumber);
+            queries.add(LAST_DIGITS_CARD_NUMBER_FIELD + "=" + URLEncoder.encode(lastDigitsCardNumber, StandardCharsets.UTF_8));
         }
         if (isSet(paymentStates)) {
-            queries.add(PAYMENT_STATES_FIELD + "=" + paymentStates.getRawString());
+            queries.add(PAYMENT_STATES_FIELD + "=" + URLEncoder.encode(paymentStates.getRawString(), StandardCharsets.UTF_8));
         }
         if (refundStates != null && refundStates.isNotEmpty()) {
-            queries.add(REFUND_STATES_FIELD + "=" + refundStates.getRawString());
+            queries.add(REFUND_STATES_FIELD + "=" + URLEncoder.encode(refundStates.getRawString(), StandardCharsets.UTF_8));
         }
         if (isSet(disputeStates)) {
-            queries.add(DISPUTE_STATES_FIELD + "=" + disputeStates.getRawString());
+            queries.add(DISPUTE_STATES_FIELD + "=" + URLEncoder.encode(disputeStates.getRawString(), StandardCharsets.UTF_8));
         }
         if (cardBrands != null && cardBrands.isNotEmpty()) {
-            queries.add(CARD_BRAND_FIELD + "=" + cardBrands.getRawString());
+            queries.add(CARD_BRAND_FIELD + "=" + URLEncoder.encode(cardBrands.getRawString(), StandardCharsets.UTF_8));
         }
         if (isNotBlank(state)) {
-            queries.add(STATE_FIELD + "=" + state);
+            queries.add(STATE_FIELD + "=" + URLEncoder.encode(state, StandardCharsets.UTF_8));
         }
         if (transactionType != null) {
-            queries.add(TRANSACTION_TYPE_FIELD + "=" + transactionType);
+            queries.add(TRANSACTION_TYPE_FIELD + "=" + URLEncoder.encode(String.valueOf(transactionType), StandardCharsets.UTF_8) );
         }
         if (isNotBlank(gatewayPayoutId)) {
-            queries.add(GATEWAY_PAYOUT_ID + "=" + gatewayPayoutId);
+            queries.add(GATEWAY_PAYOUT_ID + "=" + URLEncoder.encode(gatewayPayoutId, StandardCharsets.UTF_8));
         }
         if (isNotBlank(fromSettledDate)) {
-            queries.add(FROM_SETTLED_DATE_FIELD + "=" + fromSettledDate);
+            queries.add(FROM_SETTLED_DATE_FIELD + "=" + URLEncoder.encode(fromSettledDate, StandardCharsets.UTF_8));
         }
         if (isNotBlank(toSettledDate)) {
-            queries.add(TO_SETTLED_DATE_FIELD + "=" + toSettledDate);
+            queries.add(TO_SETTLED_DATE_FIELD + "=" + URLEncoder.encode(toSettledDate, StandardCharsets.UTF_8));
         }
         if (isNotBlank(metadataValue)) {
-            queries.add(METADATA_VALUE + "=" + metadataValue);
+            queries.add(METADATA_VALUE + "=" + URLEncoder.encode(metadataValue, StandardCharsets.UTF_8));
         }
         if (isNotBlank(agreementId)) {
-            queries.add(AGREEMENT_ID_FIELD + "=" + agreementId);
+            queries.add(AGREEMENT_ID_FIELD + "=" + URLEncoder.encode(agreementId, StandardCharsets.UTF_8));
         }
         queries.add("page=" + forPage);
         queries.add("display_size=" + getDisplaySize());
-        
-        return encodeParam(String.join("&", queries));
+        return String.join("&", queries);
     }
 
     public Long getOffset() {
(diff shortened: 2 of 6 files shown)


src/main/java/uk/gov/pay/ledger/util/pagination/PaginationBuilder.java
```diff
@@ -122,6 +124,7 @@ private void buildLinks(long lastPage) {
     }
 
     private String uriWithParams(String queryParams) {
+        queryParams = URLEncoder.encode(queryParams, StandardCharsets.UTF_8).replace("%3D", "=").replace("%26","&");
```
alexbishop1, line 127: This is a bit 😬

It looks like the `queryParams` string comes from the earlier `searchParams.buildQueryParamString(…)` call, where `searchParams` is a subclass of `SearchParams`, which is a rich object that knows the keys and values for each parameter.

So could we instead modify this `buildQueryParamString(…)` method to encode the values as it appends them to string (which, honestly, it feels like it should have been doing already and I’m not really sure how we survived this long without it)?

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
```diff
@@ -138,4 +141,27 @@ public void shouldNotGenerateNextLink_whenLimitTotalIsSetAndCountIsLessThanPageS
...
\ No newline at end of file
+    
+    @Test
+    public void shouldGenerateSelfLinkWithOnlyPrescribedSpecialCharactersURLEncoded(){
+        Map<String, Object> valuesForReferenceParameter = Map.of(
+                "{ref", "reference=%7Bref",
+                "[ref", "reference=%5Bref",
+                "r{e{f}{", "reference=r%7Be%7Bf%7D%7B",
+                "ref=", "reference=ref=",
+                "ref&", "reference=ref&",
```
alexbishop1, line 152: Having a test for a value like `foo=bar&baz=quux` (as a single value) and checking it’s not tokenised into multiple values would be useful

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
```diff
@@ -138,4 +141,27 @@ public void shouldNotGenerateNextLink_whenLimitTotalIsSetAndCountIsLessThanPageS
...
+                "ref@", "reference=ref%40"
```
alexbishop1, line 153: A test with a space would be good, given they can become `%20` or `+` depending on criteria I cannot quite remember

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
```diff
@@ -138,4 +141,27 @@ public void shouldNotGenerateNextLink_whenLimitTotalIsSetAndCountIsLessThanPageS
...
+        );
+        for (Map.Entry<String, Object> entry : valuesForReferenceParameter.entrySet()) {
```
alexbishop1, line 155: We should use [parameterised tests](https://junit.org/junit5/docs/current/user-guide/#writing-tests-parameterized-tests) to check multiple inputs

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
```diff
@@ -138,4 +141,27 @@ public void shouldNotGenerateNextLink_whenLimitTotalIsSetAndCountIsLessThanPageS
...
+            String key = entry.getKey();
+            Object value = entry.getValue();
+            transactionSearchParams.setReference(key);
+            PaginationBuilder builder = new PaginationBuilder(transactionSearchParams, mockedUriInfo)
+                    .withTotalCount(120L)
+                    .withCount(9L)
+                    .buildResponse();
+            assertThat(builder.getSelfLink().toString().contains(value.toString()), is(true));
+            assertDoesNotThrow(builder::getSelfLink);
```
alexbishop1, line 164: I think it would be good to also check that the URL ends up being what we expect (since it’s possible for this to fail in a way that gives us a valid but incorrect URL)

alexbishop1 (changes requested):
Good work tracking down roughly where this goes wrong and finding where it needs to be fixed. I’ve suggested fixing this in a different place.

We’ll need to test carefully to make sure this is behaving how we want: the fact that it’s difficult to distinguish between URL-encoded and non-encoded strings makes this a risky area to change.

src/main/java/uk/gov/pay/ledger/util/pagination/PaginationBuilder.java
hjvoid: That makes a lot more sense. I'd had a look at `buildQueryParamString(…)` but wasn't entirely sure of its function (just returns a Long ?).

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
hjvoid: Gotcha 👌!

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
hjvoid: Cool, I'll add that in too 👍

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
hjvoid: Only just noticed this in the [java coding standards](https://pay-team-manual.cloudapps.digital/manual/coding-standards/java-coding-standards.html) manual page 🤦‍♂️. I'll redo this.

src/test/java/uk/gov/pay/ledger/util/pagination/PaginationBuilderTest.java
hjvoid: I was wondering if it's better to use contains() or equals() here? I guess equals() is less likely to deliver false positives?

alexbishop1: `equals(…)` sounds better. You could do something like only look at the substring after the first `?` to reduce the clutter.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -506,9 +506,8 @@ public String buildQueryParamString(Long forPage) {
         }
         queries.add("page=" + forPage);
         queries.add("display_size=" + getDisplaySize());
-
-
-        return String.join("&", queries);
+        
+        return encodeParam(String.join("&", queries));
```
alexbishop1, line 510: This seems a little convoluted… the same thing could be achieved  (at the cost) of modifying more lines by doing:

```java
if (isNotBlank(toSettledDate)) {
    queries.add(TO_SETTLED_DATE_FIELD + "=" + URLEncoder.encode(toSettledDate, UTF-8));
}
if (isNotBlank(metadataValue)) {
    queries.add(METADATA_VALUE + "=" + URLEncoder.encode(metadataValue, UTF-8));
}
```

etc.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsTest.java
```diff
@@ -132,9 +135,9 @@ public void getsEmptyQueryParamStringWhenEmptyFromDate() {
     }
 
     @Test
-    public void getsQueryParamStringWhenNotEmptyFromDate() {
+    public void getsURLEncodedQueryParamStringWhenNotEmptyFromDate() {
         transactionSearchParams.setFromDate("2018-09-22T10:14:16.067Z");
-        assertThat(transactionSearchParams.buildQueryParamString(1L), containsString("from_date=2018-09-22T10:14:16.067Z"));
+        assertThat(transactionSearchParams.buildQueryParamString(1L), containsString("from_date=2018-09-22T10%3A14%3A16.067Z"));
```
alexbishop1, line 140: This is a behaviour change so make sure it’s tested really well.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsTest.java
```diff
@@ -170,4 +173,12 @@ public void shouldApplyToSettledDateCorrectly() {
         assertThat(transactionSearchParams.getQueryMap().get("to_settled_date"), is(ZonedDateTime.parse("2020-09-27T00:00:00.000Z")));
         assertThat(transactionSearchParams.buildQueryParamString(1L), containsString("to_settled_date=2020-09-26"));
     }
-}
\ No newline at end of file
+
+    @ParameterizedTest
+    @CsvSource({"?, %3F", "{ , %7B", "[ , %5B", "f{o{o}{ , f%7Bo%7Bo%7D%7B", "foo&, foo&", "foo@ , foo%40", "foo@@ , foo%40%40", "foo=bar&baz=quux , foo=bar&baz=quux", "foo bar , foo+bar"})
```
alexbishop1, line 178: Having each parameter on a separate line would make this easier to follow.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsTest.java
```diff
@@ -170,4 +173,12 @@ public void shouldApplyToSettledDateCorrectly() {
...
+    public void shouldUrlEncodeOnlyPrescribedSpecialCharactersOnly(String key, String value){
+        transactionSearchParams.setReference(key);
+        String expectedSelfLink = String.format("reference=%s&page=1&display_size=500", value);
```
alexbishop1, line 181: I almost missed the `%s` here: I string concatenation would actually be clearly for this specific line.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsTest.java
```diff
@@ -170,4 +173,12 @@ public void shouldApplyToSettledDateCorrectly() {
...
+        Assertions.assertEquals(transactionSearchParams.buildQueryParamString(1L), expectedSelfLink);
```
alexbishop1, line 182: We usually static import `Assertions.assertEquals` so you can just do `assertEquals(…)`.

alexbishop1 (changes requested):
Getting there! Suggested a few changes.

Make sure this is tested really well. I’d start selfservice and ledger locally, insert some payments into the database, and do some tests from the local admin tool to make sure it’s working as expected. You could even stick some temporary logging into ledger to assure yourself that the various values are what you expect at various points (for example, unencoded at the start of `TransactionSearchParams.buildQueryParamString(…)` but encoded at the end).

hjvoid:
I've updated the [ticket](https://payments-platform.atlassian.net/browse/PP-11612) to give a lowdown of what I've done regarding testing this. If there's anything I've missed there though do let me know.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -447,67 +449,65 @@ public String buildQueryParamString(Long forPage) {
...
+            queries.add(FROM_DATE_FIELD + "=" + URLEncoder.encode(fromDate, StandardCharsets.UTF_8));
         }
         if (isNotBlank(toDate)) {
-            queries.add(TO_DATE_FIELD + "=" + toDate);
+            queries.add(TO_DATE_FIELD + "=" + URLEncoder.encode(toDate, StandardCharsets.UTF_8));
         }
-
         if (isNotBlank(email)) {
-            queries.add(EMAIL_FIELD + "=" + email);
+            queries.add(EMAIL_FIELD + "=" + URLEncoder.encode(email, StandardCharsets.UTF_8));
```
alexbishop1, line 461: Can you do a manual test that searching for an email address works? I worry about the `@` sign!

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -447,67 +449,65 @@ public String buildQueryParamString(Long forPage) {
...
-            queries.add(FIRST_DIGITS_CARD_NUMBER_FIELD + "=" + firstDigitsCardNumber);
+            queries.add(FIRST_DIGITS_CARD_NUMBER_FIELD + "=" + URLEncoder.encode(firstDigitsCardNumber, StandardCharsets.UTF_8));
         }
         if (isNotBlank(lastDigitsCardNumber)) {
-            queries.add(LAST_DIGITS_CARD_NUMBER_FIELD + "=" + lastDigitsCardNumber);
+            queries.add(LAST_DIGITS_CARD_NUMBER_FIELD + "=" + URLEncoder.encode(lastDigitsCardNumber, StandardCharsets.UTF_8));
         }
         if (isSet(paymentStates)) {
-            queries.add(PAYMENT_STATES_FIELD + "=" + paymentStates.getRawString());
+            queries.add(PAYMENT_STATES_FIELD + "=" + URLEncoder.encode(paymentStates.getRawString(), StandardCharsets.UTF_8));
```
alexbishop1, line 476: Again, can you manually test this still works? You can select multiple payment states on the transactions page, so if you have a few transactions in various states you should be able to select various states and check it’s returning the expected results.

src/main/java/uk/gov/pay/ledger/util/pagination/PaginationBuilder.java
```diff
@@ -7,6 +7,8 @@
 
 import javax.ws.rs.core.UriInfo;
 import java.net.URI;
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
```
alexbishop1, line 11: Again, I think these aren’t needed any more.

src/test/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParamsTest.java
```diff
@@ -170,4 +173,21 @@ public void shouldApplyToSettledDateCorrectly() {
...
+    @ParameterizedTest
+    @CsvSource({"?, %3F", 
+            "{ , %7B", 
+            "[ , %5B", 
+            "f{o{o}{ , f%7Bo%7Bo%7D%7B", 
+            "foo&, foo%26", 
+            "foo@ , foo%40", 
+            "foo@@ , foo%40%40", 
+            "foo=bar&baz=quux , foo%3Dbar%26baz%3Dquux", 
+            "foo bar , foo+bar"
```
alexbishop1, line 186: Can we also have one where we’re searching for a literal `%` sign (i.e. double-encoding)? And can you test that manually as well please?

alexbishop1 (changes requested):
This looks good — just a few minor comments. I’ve asked for some manual testing for some scenarios — this isn’t to be annoying, it’s because this is delicate and I’m scared about breaking something! As well as using the admin tool, could you also do the same tests using public API?

I think it would be good to put some of these manual tests into code. We could have an an integration test that starts by making an HTTP request at the resource level and checks we get the expected the results. Can you look at adding one? We may have similar already. Let me know if you want some help figuring it out.

We might also want an end-to-end test or maybe a Pact test that starts from selfservice or public API and calls ledger (so we can be sure the encoding on the client-side is what ledger expects). We can think about that after this PR is ready though.

src/main/java/uk/gov/pay/ledger/common/search/SearchParams.java
```diff
@@ -2,6 +2,9 @@
 
 import uk.gov.pay.ledger.util.CommaDelimitedSetParameter;
 
+import java.net.URLEncoder;
+import java.nio.charset.StandardCharsets;
```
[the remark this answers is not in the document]
hjvoid: Not the first time I've missed this. Realised that my IDE wasn't configured to warn me about unused imports. I've changed the settings now so hopefully will pick up on these before pushing in the future.

src/main/java/uk/gov/pay/ledger/util/pagination/PaginationBuilder.java
hjvoid: As above, IDE settings now changed pick this up in the future.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
```diff
@@ -447,67 +449,65 @@ public String buildQueryParamString(Long forPage) {
         List<String> queries = new ArrayList<>();
 
         if (accountIds != null && !accountIds.isEmpty()) {
-            queries.add(GATEWAY_ACCOUNT_EXTERNAL_FIELD + "=" + String.join(",", accountIds));
+            queries.add(GATEWAY_ACCOUNT_EXTERNAL_FIELD + "=" + URLEncoder.encode(String.join(",", accountIds), StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
hjvoid: Will do, and will add to testing notes on the ticket.

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
hjvoid: As above, will add this to the manual testing notes👌

src/main/java/uk/gov/pay/ledger/transaction/search/common/TransactionSearchParams.java
hjvoid: 👌

hjvoid:
Thanks Alex! Will add add those manual tests to code as suggested in addition to adding an integration test. Will have a crack at the latter but if I get stuck will look for someone to pair with. 

> We might also want an end-to-end test or maybe a Pact test that starts from selfservice or public API and calls ledger (so we can be sure the encoding on the client-side is what ledger expects). We can think about that after this PR is ready though.

Would definitely benefit from some guidance on this 👆 as Pact tests are an area I'm pretty unfamiliar with.

src/test/java/uk/gov/pay/ledger/transaction/resource/TransactionResourceIT.java
```diff
@@ -190,6 +191,84 @@ public void shouldReturnTransactionSearchResponseObjectForNonExistentTransaction
...
+                .statusCode(Response.Status.OK.getStatusCode())
+                .contentType(JSON)
+                .body("results[0].reference", is(rawValue));
+    }
+    
+    @Test
+    public void shouldReturnATransactionSearchResponseObjectBetweenTwoDates() {
+        var fromDate = ZonedDateTime.parse("2023-11-01T00:00:00.000Z");
+        var toDate = ZonedDateTime.parse("2023-12-01T00:00:00.000Z");
+        var dateBetween = ZonedDateTime.parse("2023-11-15T14:52:07.073Z");
```
alexbishop1, line 231: I think these should all work as `Instant`s (we generally prefer `Instant`s to `ZonedDateTime`s because time zones make things complicated).

alexbishop1 (changes requested):
This looks good! Just one thing about using `Instant` rather than `ZonedDateTime` in one of the tests.

alexbishop1 (approved):

PR #3105, opened 2024-01-02
PP-12015: Add GHA to check docker sources are manifests

jfharden:
WHAT

Use the reusable workflow in pay-ci to ensure all FROM lines in the Dockerfile are pointing to manifest files


jfharden:
This repo isn't one that uses manifests (until we get to java 21) so we don't want this for now

PR #3123, opened 2024-01-12
bau: Ensure gateway account id query params are unique

oswaldquek:
Despite the class name, the fact that account ids might not be unique is bugging me 😂 Why didn't I make [getParameters return a Set](https://github.com/alphagov/pay-ledger/blob/master/src/main/java/uk/gov/pay/ledger/util/CommaDelimitedSetParameter.java#L27)? I tried, and it's [not worth the effort](https://github.com/alphagov/pay-ledger/pull/3119).


stephencdaly (approved):
Seems sensible enough. Is it worth adding a unit test?

PR #3411, opened 2024-06-11
PP-12729: Upgrade to use Java 21

kerrr:
We are upgrading to Java 21 and also switching to use graviton on AWS.

Update:
* POM file to target release 21 of java
* Dockerfile to use java 21 arm builds for mavin and release.
* Use Java 21 when setting up CodeQL environment
* Use Java 21 when setting up test environments
* Also check docker images are manifests

Due to the change of architecture, ZonedDateTime precision has been increased. This caused some tests to fail. These have been fixed by truncating to microseconds.


pom.xml
```diff
@@ -322,7 +322,7 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.13.0</version>
                 <configuration>
-                    <release>11</release>
+                    <release>21</release>
```
kbottla, line 325: Upgrading pact maven plugin will fix issues with pact publishing job as it brings in missing library

```
 <groupId>au.com.dius.pact.provider</groupId>
                <artifactId>maven</artifactId>
                <version>4.6.10</version>
```

We can also remove the below libraries. These are brought in by `pay-java-commons`
`pact-jvm-provider-junit_2.12`
`pact-jvm-consumer-junit_2.12`

kbottla: Also need to update `.java-version` file

src/test/java/uk/gov/pay/ledger/event/dao/EventDaoIT.java
```diff
@@ -15,6 +15,7 @@
 import java.io.IOException;
 import java.sql.Timestamp;
 import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
```
kbottla, line 18: Unused import?

pom.xml
```diff
@@ -322,7 +310,7 @@
...
```
kbottla, line 313: .java-version still says 11

kerrr: odd 👀

kerrr: It's there now.

src/test/java/uk/gov/pay/ledger/event/dao/EventDaoIT.java
kerrr: Thanks

kbottla (approved):

PR #3456, opened 2024-07-04
PP-12841 add new fields to EventTicker class

nlsteers:
### WHAT

add the following fields to the EventTicker class for consumption via the live payments dashboard:

- service_external_id
- wallet_type
- source
- is_moto
- is_recurring

add new query parameter to `/ticker` resource for filtering events by salient event type


sfount (commented):
🍬 functionality lines up with the ticket and code looks sensible to me. If its too gnarly to provide a list of events from the consumer the event filtering could also be defined by Ledger with the consumer just providing a boolean; asking for everything or a set filter (successful payment confirm events or similar).

LGTM, great work 👍

DomBelcher (approved):

PR #3635, opened 2024-11-12
PP-13344 return empty string

SandorArpa:
When a payment in ledger has its requires_3ds property false the value of the ‘3-D Secure Required’ column is an empty string

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
@@ -127,7 +127,8 @@ public class CsvTransactionFactory {
                     result.putAll(getFeeBreakdown(transactionEntity, feeBreakdownNode));
                 }
 
-                result.put(FIELD_3D_SECURE_REQUIRED, safeGetAsBoolean(transactionDetails, "requires_3ds", null));
+                Boolean requires3ds = safeGetAsBoolean(transactionDetails, "requires_3ds", null);
+                result.put(FIELD_3D_SECURE_REQUIRED, Boolean.TRUE.equals(requires3ds) ? true : null);
 
                 Optional<Map<String, Object>> externalMetadata = getExternalMetadata(transactionEntity.getTransactionDetails());
 
--- a/src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
+++ b/src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
@@ -376,7 +376,7 @@ class CsvTransactionFactoryTest {
     }
 
     @Test
-    void toMapShouldIncludeShouldInclude3DSecureRequired() {
+    void toMapShouldInclude3DSecureRequiredWhenTrue() {
         var transactionEntity = transactionFixture
                 .withTotalAmount(123L)
                 .withVersion3ds("2.0.1")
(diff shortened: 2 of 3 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
```diff
@@ -127,7 +127,11 @@ public Map<String, Object> toMap(TransactionEntity transactionEntity) {
                     result.putAll(getFeeBreakdown(transactionEntity, feeBreakdownNode));
                 }
 
-                result.put(FIELD_3D_SECURE_REQUIRED, safeGetAsBoolean(transactionDetails, "requires_3ds", null));
+                Boolean requires3ds = safeGetAsBoolean(transactionDetails, "requires_3ds", null);
+                if (Boolean.FALSE.equals(requires3ds)) {
+                    requires3ds = null;
+                }
+                result.put(FIELD_3D_SECURE_REQUIRED, requires3ds);
```
alexbishop1, line 134: Might this be more concise?

```java
Boolean required3ds = safeGetAsBoolean(transactionDetails, "requires_3ds", null);
result.put(FIELD_3D_SECURE_REQUIRED, required3ds ? true : null);
```

src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
```diff
@@ -403,6 +403,46 @@ void toMapShouldIncludeShouldInclude3DSecureRequired() {
...
+                .withCreatedDate(ZonedDateTime.parse("2018-03-12T16:25:01.123456Z"))
+                .withCorporateCardSurcharge(23)
+                .withCardBrandLabel("Visa")
+                .withDefaultCardDetails()
+                .withFee(5L)
+                .withTotalAmount(123L)
+                .withRequires3ds(false)
+                .withVersion3ds("2.0.1")
+                .withDefaultTransactionDetails()
+                .toEntity();
```
alexbishop1, line 423: Looks like this could be a bit less verbose by following the pattern in `toMapShouldIncludeShouldInclude3DSecureRequired()`.

src/test/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactoryTest.java
```diff
@@ -403,6 +403,46 @@ void toMapShouldIncludeShouldInclude3DSecureRequired() {
...
+        assertThat(csvDataMap.get("Net"), is("0.00"));
+        assertThat(csvDataMap.get("Error Code"), is("P0010"));
+        assertThat(csvDataMap.get("Error Message"), is("Payment method rejected"));
+        assertThat(csvDataMap.get("Date Created"), is("12 Mar 2018"));
+        assertThat(csvDataMap.get("Time Created"), is("16:25:01"));
+        assertThat(csvDataMap.get("Corporate Card Surcharge"), is("0.23"));
+        assertThat(csvDataMap.get("Total Amount"), is("0.00"));
+        assertThat(csvDataMap.get("Payment Provider"), is("sandbox"));
+        assertThat(csvDataMap.get("3-D Secure Required"), is(nullValue()));
+    }
```
alexbishop1, line 444: Perhaps add a variant of this where it’s explicitly null and also rename `toMapShouldIncludeShouldInclude3DSecureRequired()` (which is wrong already!) to `toMapShouldInclude3DSecureRequiredWhenTrue()` or something.

alexbishop1 (changes requested):
Looks like it should work: a couple of small suggestions.

src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
alexbishop1: Actually, don’t do that! It will blow up with a `NullPointerException` when `required3ds` is null!

I think the best we can do is probably something like this, which is similar to what you originally had:

```java
if (Boolean.TRUE.equals(safeGetAsBoolean(transactionDetails, "requires_3ds", null))) {
    result.put(FIELD_3D_SECURE_REQUIRED, true);
} else {
    result.put(FIELD_3D_SECURE_REQUIRED, null);
}
```

It’s a shame we don’t live in a world where [JEP 455](https://openjdk.org/jeps/455) has shipped because then we could do this:

```java
result.put(FIELD_3D_SECURE_REQUIRED, switch(safeGetAsBoolean(transactionDetails, "requires_3ds", null)) {
    case true -> true;
    case null, default -> null;
};

SandorArpa: I didn't do that. I did this:
Boolean requires3ds = safeGetAsBoolean(transactionDetails, "requires_3ds", null);
result.put(FIELD_3D_SECURE_REQUIRED, Boolean.TRUE.equals(requires3ds) ? true : null);

alexbishop1 (approved):

PR #3649, opened 2024-11-21
Add Exemption object to Ledger Specification

marcotranchino:
With this change, we are starting to introduce the exemption object in the Ledger specification so that it will be added to the TransactionView.

If an exemption was requested, the new object will be set with the property `requested` set to true. If we know that an exemption was not requested, then the new object will be set with the property `requested` set to false.

Further information in the JIRA ticket below.

https://payments-platform.atlassian.net/browse/PP-13332


SandorArpa (changes requested):
Can we add a couple of tests?

SandorArpa (changes requested):
Nice, however it'd be beneficial if we added at least a couple of tests to `TransactionServiceTest` to prove that the underlying logic adds/does not add the Exemption object.
And perhaps one to `TransactionResourceIT` to prove that the Json outcome is rendered as intended. This will also be useful when we will extend it to accommodate additional fields/objects.

SandorArpa (approved):
LGTM

PR #3714, opened 2025-01-20
PP-13515 refactor queue msg pact tests

SandorArpa:
Current setup checks if the type of the JsonElement of the Json payload coming from the provider (connector) is matching the type of the JsonElement of the Json payload of what the consumer (ledger) expects. This will match `honoured` to `HONOURED` or `EXEMPTION_HONOURED` to `HONOURED`.

Refactor `QueueEventFixtureUtil.getNestedPact()` to use strict, value based matching of strings. Also refactor existing tests to match values generated by `connector`


hjvoid (approved):
LGTM😎

PR #3791, opened 2025-03-31
PP-13654 upgrade modules to dropwizard 4

SandorArpa:
Upgrade to use Dropwizard 4.
Also upgrade, what we can to Java Jakarta EE.


pom.xml
```diff
@@ -159,39 +141,38 @@
...
+            <artifactId>queue-dropwizard-4</artifactId>
             <version>${pay-java-commons.version}</version>
         </dependency>
         <dependency>
-            <groupId>org.glassfish.jersey.bundles.repackaged</groupId>
-            <artifactId>jersey-guava</artifactId>
-            <version>2.25.1</version>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>33.4.6-jre</version>
```
alexbishop1, line 170: Dropwizard Dependencies already provides Guava so we should just be able to have an unversioned dependency in the unversioned dependencies section

alexbishop1: Seems like this is still here…

alexbishop1 (changes requested):

pom.xml
SandorArpa: 🤦 🤦 🤦 
I'll hand in my resignation tomorrow

alexbishop1 (approved):

PR #3806, opened 2025-04-11
PP-13622 Fix incorrect negation for won disputes in CSV files

marcotranchino:
With this change, we are fixing a bug for which disputed amounts are shown as negative numbers in the CSV files.

The logic will now consider also the state of the dispute and present the value as positive when the dispute has been won by the service.

Further information in Jira.

https://payments-platform.atlassian.net/browse/PP-13622

### DIFF
--- a/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
+++ b/src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
@@ -31,6 +31,7 @@ import static org.apache.commons.lang3.StringUtils.replaceChars;
 import static org.apache.commons.text.WordUtils.capitalizeFully;
 import static uk.gov.pay.ledger.transaction.model.TransactionType.DISPUTE;
 import static uk.gov.pay.ledger.transaction.model.TransactionType.REFUND;
+import static uk.gov.pay.ledger.transaction.state.TransactionState.WON;
 import static uk.gov.pay.ledger.util.JsonParser.safeGetAsBoolean;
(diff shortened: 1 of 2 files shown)


src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
```diff
@@ -154,7 +155,12 @@ public Map<String, Object> toMap(TransactionEntity transactionEntity) {
                         getPaymentTransactionAttributes(transactionEntity, transactionDetails.get("payment_details"))
                 );
                 result.put(FIELD_GOVUK_PAYMENT_ID, transactionEntity.getParentExternalId());
-                result.put(FIELD_AMOUNT, penceToCurrency(transactionEntity.getAmount() * -1));
+                boolean isWonDispute = transactionEntity.getState() == WON;
```
SandorArpa, line 158: You can just do:
`if(transactionEntity.getState() == WON)`.  The variable `isWonDispute` is used only for the if statement.

SandorArpa (changes requested):

src/main/java/uk/gov/pay/ledger/transaction/model/CsvTransactionFactory.java
marcotranchino: Ah good point, thanks!

marcotranchino: Amended, thanks!

SandorArpa (approved):
LOKTM

PR #3937, opened 2025-08-19
PP-13991 refactor to use vanilla SentryAppenderFactory

SandorArpa:
import `SentryAppenderFactory` from dropwizard-sentry.


hjvoid (approved):
LGTM

PR #3985, opened 2025-10-01
PP-14434 return rcp reason property in payment response

SandorArpa:
If `agreemnt_payment_type` exists in payment details, Ledger returns reason property in payment response objects


AbdirizakIdris (approved):
Looks good to me

